Issue #3, opened 2025-04-24
[bug] Slient failure on initialize


mayocream:
related: https://github.com/tauri-apps/tauri/discussions/11831

Currently, the initialize function in Rust fails without any notice, which causes the splash screen to show infinitely.

Phundrak:
Possibly related to this issue, when I try to launch koharu, I end up with a blank window, as seen below.

![Image](https://github.com/user-attachments/assets/6f5bc1ee-be7e-4f8c-a8c7-f34aceb05266)

This is what I see in the terminal when launching the program in development mode:

```
phundrak@marpa$ dev
$ tauri dev
     Running BeforeDevCommand (`bun run dev`)
$ cd next && bun run dev
$ next dev -p 9000 --turbopack 
   ▲ Next.js 15.3.1 (Turbopack)
   - Local:        http://localhost:9000
   - Network:      http://192.168.0.86:9000

 ✓ Starting...
 ✓ Ready in 558ms
     Running DevCommand (`cargo  run --no-default-features --color always --`)
        Info Watching /home/phundrak/fromGIT/koharu/src-tauri for changes...
        Info Watching /home/phundrak/fromGIT/koharu/src-tauri for changes...
        Info Watching /home/phundrak/fromGIT/koharu/yolo-v8 for changes...
        Info Watching /home/phundrak/fromGIT/koharu/comic-text-detector for changes...
        Info Watching /home/phundrak/fromGIT/koharu/manga-ocr for changes...
        Info Watching /home/phundrak/fromGIT/koharu/lama for changes...
   Compiling koharu v0.1.0 (/home/phundrak/fromGIT/koharu/src-tauri)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 8.16s
     Running `/home/phundrak/fromGIT/koharu/target/debug/koharu`
Please make sure to register the `fs` plugin before the `persisted-scope` plugin!
```

I can open the Next.js link in the logs, which gives me this page.

![Image](https://github.com/user-attachments/assets/fe79e000-4a62-4faa-8b70-b13c8bfd88e4)

However, when clicking on the top-left icon, I get the following errors:

![Image](https://github.com/user-attachments/assets/944b1a15-2955-48bb-a982-25e9344b8124)

Firefox devtools error message:
```
Uncaught (in promise) TypeError: window.__TAURI_INTERNALS__ is undefined
    invoke core.js:188
    open index.js:62
    handleOpenFile topbar.tsx:15
[...]

PR #5, opened 2025-04-28
feat: add reproducible environment with Nix flakes

Phundrak:
This commit adds a flakes.nix file that declares a reproducible environment. It is usable with the Nix[1] package manager (NixOS not required) using the Flakes feature[2]. It also uses the devenv[3] module for an easy way to declare installed packages, commands, tests, tasks, and git hooks.

This reproducible environment provides a shell with all required tools available for development and building the project.

[1] https://nixos.org/
[2] https://nixos.wiki/wiki/flakes
[3] https://devenv.sh/


Phundrak:
This is only a draft PR, as I cannot fully test this due to koharu not launching properly on my machine. This is possibly related to #3.

PR #158, opened 2026-01-30
Fix vertical column pitch

MARJORIESHA-pBAD:
Some fonts report large ascent/descent, which inflates line_height. When line_height is used as the vertical column pitch, columns become too wide and the ink occupies only a small portion of the box. This change, therefore, uses font_size for vertical column pitch and bounds width, and keeps ink bounds for clipping safety.

Demonstration of issue:
Screenshot 2026-01-30 171714
Screenshot 2026-01-30 165108
Screenshot 2026-01-30 165058
Screenshot 2026-01-30 162307


mayocream (approved):
LGTM

PR #198, opened 2026-03-05
feat: add Spanish translations and update language options

karrot0:
Added spanish translation and updated other files to include the russian and spanish option


mayocream:
Thank you @karrot0!

PR #214, opened 2026-03-09
LLM API

karrot0:
This pr adds LLM Api's to the model selector and add api key fields to the settings page.

image

image

image

image

API keys are stored in OS Keychain for safety. For the model selection I tried to find the cheapest inference and translation quality > everything else.


ui/app/providers.tsx
```diff
@@ -31,6 +33,22 @@ export function Providers({ children }: { children: ReactNode }) {
     ensureDownloadSubscribed()
   }, [ensureDownloadSubscribed])
 
+  useEffect(() => {
+    if (!isTauri()) return
```
mayocream: Since the API key is stored in the keyring, it should work for both the web and the Tauri environments. (?)

koharu-runtime/src/dylib.rs
```diff
@@ -72,6 +72,17 @@ pub const PACKAGES: &[&str] = &[
...
+#[cfg(target_os = "windows")]
+const COMPAT_ALIASES: &[(&str, &str)] = &[
+    #[cfg(feature = "cuda")]
+    ("cudart64_13.dll", "cudart64_12.dll"),
+    #[cfg(feature = "cuda")]
+    ("cublasLt64_13.dll", "cublasLt64_12.dll"),
+    #[cfg(feature = "cuda")]
+    ("cublas64_13.dll", "cublas64_12.dll"),
+];
+
```
mayocream: `cudarc` does support searching CUDA 13, ref: https://github.com/chelsea0x3b/cudarc/blob/53d92abd7a3160ee8eaf3d679dbee1508572fe94/src/lib.rs#L203

We use `cuda-version-from-build-system` feature flag, and `cudarc` will use your system's CUDA version (via `nvcc`) to render `"{DLL_PREFIX}{lib_name}{pointer_width}_{major}{DLL_SUFFIX}"`, so if your system installs CUDA 12, it will fail to find your DLLs.

Let's not change this in this PR; for your dev env, you can use `cuda-12090` flag to replace `cuda-version-from-build-system`. Or update to CUDA 13.1.

PR #275, opened 2026-03-25
New features and fix canvas and other... 

spgsroot:
I continue to gradually find problems in my translation automation path :D 

New features                                                                                     
- Magic Eraser tool (M) - free-form inpaint on any canvas area via LaMa, no prior detection required. New backend endpoint POST                             /documents/{id}/inpaint-free                                                                                                                              
- Undo/Redo (Ctrl+Z / Ctrl+Shift+Z, Cmd on Mac) - for text block edits, magic eraser, repair brush, and brush strokes                                       - Ctrl+Click multi-select - select multiple text blocks on canvas and in sidebar, batch edit color/font/size/effects/stroke                                 - Font size control - +/- stepper in Render panel (4–200px)                                                                                                 
- Original Image toggle - click in Layers panel or press O to view only the raw image                                                                       
- Sensitive Detect (Detect+) - lower thresholds (0.10 vs 0.25) to catch SFX, small text, complex backgrounds                                                - CTD fallback - when PP-DocLayoutV3 finds nothing, Comic Text Detector (YoloV5+DBNet) runs automatically                                                 
  - Region detect API - backend support for detecting text in a selected sub-region
  - Tool keyboard shortcuts - V/B/P/E/R/M for all canvas tools
  - Zoom up to 400% for detailed editing
{AF872EB3-D25B-4C2B-8B43-8C8AE63BA985}
{E14A80A1-84AB-4B05-B159-D5EAB479844B}
{253EA54F-5076-4698-A66D-21A353BC31D9}
{9525D306-0ECA-4838-AD7E-C27E0EF02F54}

  Fixes
  - Blocks snapping back on drag - split optimistic cache update (instant) from backend sync (debounced)
  - Undo closures using stale data - read from React Query cache instead of captured render state
  - Canvas state destroyed on Original Image toggle - replaced conditional unmount with CSS visibility
  - Stale currentDocumentIndex in mask/brush drawing hooks - read from store at call time
  - Tauri resize listener ignoring autoFitEnabled changes - read from store in callback
  - Unstable React keys in TextBlockAnnotations - changed to block.id
  - useEffect overwriting position during active drag - draggingRef guard
  - Undo flooding on micro-drags - single entry per drag session
  - Silent errors in mask/inpaint operations - visible via useUiErrorStore
  - Detect spinner showing on both Detect/Detect+ - track which button triggered
  - Render panel clipped with scrollbar - removed fixed height, content fits naturally

PR #281, opened 2026-03-26
feat: detect+, CTD fallback, region detect, magic eraser & LLM backend refactor

spgsroot:
## Summary
Backend changes split from #275:

**LLM Backend Refactor**
- Remove shared `Arc<LlamaBackend>` — each `Llm` instance now owns its backend
- Move PaddleOcrVl from `koharu-llm` to `koharu-ml`

**Detect+ (Sensitive Mode)**
- `DetectOptions { sensitive: bool }` with lower thresholds (0.10 vs 0.25) for SFX, small text, complex backgrounds
- Configurable min block dimensions and area filtering

**CTD Fallback**
- When PP-DocLayoutV3 finds no text blocks, Comic Text Detector runs automatically
- CTD results are converted to `TextBlock` with `detector: Some("ctd")`

**Region Detect**
- Detect text in a user-selected sub-region of the image
- Appends new blocks to existing detections, re-runs segmentation

**Magic Eraser (Free-form Inpaint)**
- `inpaint_free()` pipeline operation — inpaints arbitrary regions via LaMa without requiring text blocks

**New API Endpoints**
- `POST /documents/{id}/detect-options` — detect with sensitivity and optional region
- `POST /documents/{id}/inpaint-free` — free-form inpaint for magic eraser

## Test plan
- [ ] Verify detect with default options still works as before
- [ ] Test sensitive mode catches more text blocks
- [ ] Test CTD fallback triggers when PP-DocLayout returns empty
- [ ] Test region detect appends blocks correctly
- [ ] Test magic eraser inpaints without text block requirements
- [ ] Verify LLM loading works with per-instance backend

### DIFF
--- a/ui/components/canvas/Workspace.tsx
+++ b/ui/components/canvas/Workspace.tsx
@@ -181,14 +181,14 @@ export function Workspace() {
   })
   const { t } = useTranslation()
 
-  // Listen for Tauri resize events
+  // Listen for Tauri resize events — register once, read state from store.
   useEffect(() => {
     let unlisten: (() => void) | undefined
 
     const setupListener = async () => {
       unlisten = await listen('tauri://resize', () => {
-        const { autoFitEnabled: fitEnabled } = useEditorUiStore.getState()
-        if (currentDocument && fitEnabled) {
+        const { autoFitEnabled, totalPages } = useEditorUiStore.getState()
+        if (totalPages > 0 && autoFitEnabled) {
           fitCanvasToViewport()
         }
       })
@@ -201,7 +201,7 @@ export function Workspace() {
         unlisten()
       }
     }
-  }, [currentDocument])
+  }, [])
 
   useGesture(
     {
--- a/ui/hooks/useBrushLayerDisplay.ts
+++ b/ui/hooks/useBrushLayerDisplay.ts
@@ -38,6 +38,10 @@ export function useBrushLayerDisplay({
       canvas.height = currentDocument.height
     }
 
+    // Always clear before async load so stale content from a previous page
+    // doesn't flash when both pages share the same canvas dimensions.
+    ctx?.clearRect(0, 0, canvas.width, canvas.height)
+
     let cancelled = false
     const brushLayer = currentDocument.brushLayer
 
@@ -66,8 +70,6 @@ export function useBrushLayerDisplay({
           console.error(error)
         }
       })()
-    } else {
-      ctx?.clearRect(0, 0, canvas.width, canvas.height)
     }
 
     return () => {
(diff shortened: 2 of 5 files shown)


koharu-ml/src/facade.rs
```diff
@@ -1,17 +1,13 @@
-use std::{
-    sync::{Arc, Mutex},
-    time::Instant,
-};
+use std::{sync::Mutex, time::Instant};
 
 use anyhow::Result;
 use image::DynamicImage;
-use koharu_llm::paddleocr_vl::{self as paddleocr_vl_llm, PaddleOcrVl, PaddleOcrVlTask};
-use koharu_llm::safe::llama_backend::LlamaBackend;
```
mayocream, line 9: We need to use llama.cpp-based paddleocr-vl impl, please revert these changes.

koharu-ml/src/facade.rs
```diff
@@ -87,11 +95,16 @@ pub struct Model {
 }
 
 impl Model {
-    pub async fn new(cpu: bool, backend: Arc<LlamaBackend>) -> Result<Self> {
+    /// Access the underlying comic text detector (for segmentation in region-detect).
+    pub fn segmenter(&self) -> &ComicTextDetector {
+        &self.segmenter
+    }
```
mayocream: This is unnecessary, I think.

koharu-ml/src/facade.rs
```diff
@@ -87,11 +95,16 @@ pub struct Model {
...
+
+    pub async fn new(cpu: bool) -> Result<Self> {
         Ok(Self {
             layout_detector: PPDocLayoutV3::load(cpu).await?,
-            segmenter: ComicTextDetector::load_segmentation_only(cpu).await?,
-            ocr: Mutex::new(PaddleOcrVl::load(cpu, backend).await?),
+            segmenter: ComicTextDetector::load(cpu).await?,
```
mayocream: let's rename it to `ctd` instead of `segmenter`

koharu-pipeline/src/ops/edit.rs
```diff
@@ -598,6 +598,72 @@ pub async fn inpaint_partial(
     .await
 }
 
+/// Free-form inpaint: inpaint an arbitrary region using the mask without requiring
+/// text blocks.  Used by the Magic Eraser tool so users can erase defects, artifacts
+/// or any unwanted content anywhere on the canvas.
+#[instrument(level = "info", skip_all)]
```
mayocream: It might be confusing that there are two different inpaint brush tools. We should change the existing `inpaint_partial`'s behavior.

koharu-rpc/src/api.rs
```diff
@@ -67,6 +67,10 @@ pub fn router(resources: SharedResources, events: EventHub) -> Router {
             get(get_document_layer),
         )
         .route("/documents/{document_id}/detect", post(detect_document))
+        .route(
+            "/documents/{document_id}/detect-options",
+            post(detect_document_options),
+        )
```
mayocream: Same as `inpaint_partial`, we should add options to `detect` api, not adding a new duplciate api.

PR #290, opened 2026-03-27
ci: cache bun packages and Next.js build output

EverythingSuckz:
## Summary
- Add `actions/cache@v5` for bun package cache (`~/.bun/install/cache`) to speed up `bun install`
- Add `actions/cache@v5` for Next.js build cache (`ui/.next/cache`) to eliminate "No build cache found" warning and speed up frontend compilation

Applied to all workflows that build the frontend: `build.yml`, `release.yaml`, `publish.yml`.

## References
- https://github.com/actions/cache/blob/main/examples.md#bun
- https://nextjs.org/docs/messages/no-cache

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,17 +23,7 @@ jobs:
           method: 'network'
           sub-packages: '["nvcc", "cudart", "crt", "thrust", "nvvm", "cublas", "cublas_dev", "cufft", "cufft_dev", "curand", "curand_dev", "nvrtc", "nvrtc_dev"]'
       - uses: ilammy/msvc-dev-cmd@v1
-      - uses: actions/cache@v5
-        with:
-          path: ~/.bun/install/cache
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -23,7 +23,17 @@ jobs:
           method: 'network'
           sub-packages: '["nvcc", "cudart", "crt", "thrust", "nvvm", "cublas", "cublas_dev", "cufft", "cufft_dev", "curand", "curand_dev", "nvrtc", "nvrtc_dev"]'
       - uses: ilammy/msvc-dev-cmd@v1
+      - uses: actions/cache@v5
```
mayocream: Let's not use cache on publish & release; a fresh build is required to reduce stale cache issues.

EverythingSuckz: Yep, totally makes sense.

Issue #294, opened 2026-03-28
[Bug] Linux panic on startup – failed to load libggml-base.so


karrot0:
### Error
thread 'main' panicked at /mnt/karrot/Projects/koharu/koharu/src/app.rs:275:22:
failed to build app resources: failed to initialize llama.cpp runtime bindings

Caused by:
    0: failed to load `libggml-base.so`
    1: failed to load `libggml-base.so`
    2: libggml-base.so: cannot open shared object file: No such file or directory

### Environment
- OS: Linux (Arch Linux)
- Session: Wayland
- GPU: NVIDIA (CUDA 13.2 reported)
- Build: dev release

PR #295, opened 2026-03-28
fix(linux): load native libraryes by using explicit path to resolve missing libggml errors on Unix systems

karrot0:
image

Fixes https://github.com/mayocream/koharu/issues/294


mayocream (approved):
Thank you for catching this up!

PR #329, opened 2026-04-03
Cuda 13 0

cristian-guerrero:


koharu-runtime/src/llama.rs
```diff
@@ -154,96 +172,152 @@ impl LlamaDistribution {
...
+        // The extracted directory name is llama.cpp-<tag>
+        let source_root = build_dir.join(format!("llama.cpp-{LLAMA_CPP_TAG}"));
+        let cmake_build_dir = source_root.join("build");
+        fs::create_dir_all(&cmake_build_dir)?;
+
+        let mut cmake = Command::new("cmake");
+        cmake.current_dir(&cmake_build_dir)
+            .arg("..")
+            .arg("-DBUILD_SHARED_LIBS=ON")
+            .arg("-DGGML_CUDA=ON")
```
mayocream, line 236: Building llama.cpp against the CUDA feature will take very, very, very long; that's why we use a prebuilt release. 😿

PR #351, opened 2026-04-06
You can now see if you downloaded a llm model in the ui

louisoes05-png:
image


model-selector-download-badges.patch
```diff
@@ -0,0 +1,458 @@
...
+               ) : null}
+-              {selectedIsLoaded || llmUnloading
+-                ? t('llm.unload')
+-                : t('llm.load')}
++              {selectedIsLoaded || llmUnloading ? (
++                t('llm.unload')
++              ) : selectedRequiresDownload ? (
++                <span className='flex flex-col items-start leading-tight'>
++                  <span>
++                    {t('llm.downloadAndLoad', {
```
karrot0: Need to add this value to all translation json in ui/public/locales for translation

PR #357, opened 2026-04-06
feat(mt): DeepL/Google as koharu-llm providers

Duvet13d:
## Summary
- Add `deepl` and `google-translate` as pluggable providers in `koharu-llm` (new provider files + `mod.rs` registration).
- Route translation through the provider system when pipeline translator is `deepl`/`google-translate`.

## Why
Matches the existing provider extension pattern (OpenAI/Gemini/Claude): add provider logic in a new file and register it in `mod.rs`, without introducing a separate engine system.

## Validation
- bun cargo fmt -- --check
- bun cargo check

### DIFF
--- a/koharu-app/src/engine.rs
+++ b/koharu-app/src/engine.rs
@@ -1002,31 +1002,6 @@ inventory::submit! {
     }
 }
 
-// Aliases so users can select these in Settings → Engines without adding a new engine type.
-inventory::submit! {
-    EngineInfo {
-        id: DEEPL_ID,
-        name: "DeepL",
-        needs: &[Artifact::OcrText],
-        produces: &[Artifact::Translations],
-        load: |_res| Box::pin(async move {
-            Ok(Box::new(LlmTranslateEngine) as Box<dyn Engine>)
-        }),
-    }
-}
-
-inventory::submit! {
-    EngineInfo {
-        id: GOOGLE_TRANSLATE_ID,
-        name: "Google Cloud Translation",
-        needs: &[Artifact::OcrText],
-        produces: &[Artifact::Translations],
-        load: |_res| Box::pin(async move {
-            Ok(Box::new(LlmTranslateEngine) as Box<dyn Engine>)
-        }),
-    }
-}
-
 // --- Lama Inpainting ------------------------------------------------------
 
 struct LamaInpaintEngine(koharu_ml::lama::Lama);


koharu-llm/src/providers/deepl_mt.rs
```diff
@@ -0,0 +1,133 @@
...
+    fn translate<'a>(
+        &'a self,
+        source: &'a str,
+        target_language: Language,
+        _model: &'a str,
+        _custom_system_prompt: Option<&'a str>,
+    ) -> Pin<Box<dyn Future<Output = anyhow::Result<String>> + Send + 'a>> {
+        Box::pin(async move {
+            let root = normalize_base_url(self.base_url.as_deref());
+            let url = format!("{root}/v2/translate");
```
karrot0, line 105: API endpoint won't work for free users, will need to use https://api-free.deepl.com only for free users

Image

koharu-llm/Cargo.toml
```diff
@@ -51,6 +51,7 @@ tokio = { workspace = true }
 tracing = { workspace = true }
 tracing-subscriber = { workspace = true }
 tracing-core = "0.1"
+url = { workspace = true }
```
karrot0: used only once to encode form, might want to remove it and use the already existing dependencies like reqwest https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.form

karrot0 (changes requested):
Please fix the API Endpoint and consider using reqwest form for encoding instead of url. 

Here are pictures to show what it looks like

Image

Image

Image

koharu-app/src/engine.rs
```diff
@@ -928,16 +931,57 @@ impl Engine for LlmTranslateEngine {
...
         let block_count = page.text_blocks.len();
         async {
-            res.llm
-                .translate(
-                    &mut page,
-                    options.target_language.as_deref(),
-                    options.system_prompt.as_deref(),
-                )
-                .await
+            if translator == DEEPL_ID || translator == GOOGLE_TRANSLATE_ID {
```
mayocream, line 941: Can we put this logic into providers instead of in the engine?

karrot0 (approved):

koharu-app/src/engine.rs
Duvet13d: sure, I would try

Issue #339, opened 2026-04-05
[Feature Suggestion] Translation Glossary


karrot0:
Let me know what u think of a translation glossary for models with large context window. It would help for replacing specific words and providing more translation context for the model. Could improve terminology aswell.

We could add a tab to settings were the user can add it themselves and be saved to memory or be persistent.

lmk what you think of this

mayocream:
Reopening as the original glossary request; configurable terminology is still missing.

PR #369, opened 2026-04-06
feat: add glossary

karrot0:
### Changes:

- Implementation of https://github.com/mayocream/koharu/issues/339

- creates new tab in settings called "Glossary", which has a rows and source toggle,

- can add easily new values

- added some missing translation to all locale files

### Images:

image


mayocream:
@Map1en can you help review this? thanks!

ui/components/SettingsDialog.tsx
```diff
@@ -882,6 +893,233 @@ function AboutPane({
...
+
+  const enterSource = () => {
+    setSourceText(
+      glossary.map(({ source, target }) => `${source} = ${target}`).join('\n')
+    )
+    setSourceError(null)
+    setViewMode('source')
+  }
+
+  const applySource = () => {
```
Map1en, line 979: source text handling has some issues. If the format is not correct, entries may be removed by mistake, and the deduplication logic also seems unreliable. I suggest switching this to CSV instead. It would be more common easier to use.

Map1en (changes requested):
source text handling has some issues. If the format is not correct, entries may be removed by mistake, and the deduplication logic also seems unreliable. I suggest switching this to CSV instead. It would be more common easier to use.

PR #375, opened 2026-04-07
Improve drag-and-drop imports and show downloaded local models

louisoes05-png:
## Summary
- add app-wide drag and drop support for images and folders
- scan dropped folders recursively, skip duplicate or unreadable imports safely, and improve drop feedback
- show a Downloaded badge for cached local models and update the action text for models that still need to be fetched

## Testing
- bun x tsc --noEmit
- cargo fmt --package koharu-core --package koharu-rpc --package koharu-app
- bun run dev
- cargo check -p koharu-rpc -p koharu-core (blocked here because koharu-llm build.rs tries to download llama.cpp from GitHub)

### DIFF
--- a/koharu-rpc/src/api.rs
+++ b/koharu-rpc/src/api.rs
@@ -1664,10 +1664,9 @@ fn collect_image_entries_from_path(path: &std::path::Path) -> CollectedImportEnt
 
             if entry_metadata.is_dir() {
                 collected.skipped_count += 1;
-                collected.warnings.push(format!(
-                    "Skipped nested folder: {}",
-                    entry_path.display()
-                ));
+                collected
+                    .warnings
+                    .push(format!("Skipped nested folder: {}", entry_path.display()));
                 continue;
             }
 


koharu-rpc/src/api.rs
```diff
@@ -597,10 +599,103 @@ async fn import_documents(
...
+            warnings: Vec::new(),
+        },
+        |mut collected, path| {
+            let result = collect_image_entries_from_path(std::path::Path::new(path));
+            collected.files.extend(result.files);
+            collected.skipped_count += result.skipped_count;
+            collected.warnings.extend(result.warnings);
+            collected
+        },
+    );
```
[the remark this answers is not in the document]
liksunrice: @copilot apply changes based on [this feedback](https://github.com/mayocream/koharu/pull/375#discussion_r3043324552)

koharu-llm/src/lib.rs
```diff
@@ -52,7 +52,8 @@ pub enum ModelId {
         props(
             repo = "lmg-anon/vntl-llama3-8b-v2-gguf",
             filename = "vntl-llama3-8b-v2-hf-q8_0.gguf",
-            languages = "en-US"
+            languages = "en-US",
+            download_size = "~8 GB"
```
fffonion: consider use an easily parsable number instead of hard coding the human readable literal.

koharu-rpc/src/api.rs
```diff
@@ -1456,6 +1553,147 @@ fn binary_response(data: Vec<u8>, content_type: &str, filename: Option<String>)
...
+
+fn is_allowed_local_origin(origin: &str) -> bool {
+    origin.starts_with("http://localhost:")
+        || origin.starts_with("http://127.0.0.1:")
+        || origin == "http://tauri.localhost"
+        || origin == "https://tauri.localhost"
+        || origin == "tauri://localhost"
+}
+
+fn collect_image_entries_from_path(path: &std::path::Path) -> CollectedImportEntries {
```
fffonion, line 1595: loading recursively by default is probably not good idea for security. let's by default load top level files.
recursive loading can be triggered by for example pressing a special key (control/shift) and can be implemented in another PR.

koharu-rpc/src/api.rs
```diff
@@ -1456,6 +1553,147 @@ fn binary_response(data: Vec<u8>, content_type: &str, filename: Option<String>)
...
+fn ensure_local_import_origin(headers: &HeaderMap) -> ApiResult<()> {
+    let Some(origin) = headers.get(ORIGIN) else {
+        return Ok(());
+    };
+
+    let origin = origin
+        .to_str()
+        .map_err(|_| ApiError::bad_request("Invalid Origin header"))?;
+
+    if is_allowed_local_origin(origin) {
```
fffonion, line 1577: this should be disabled on headless mode (e.g. non-tauri), only relying on CORS is not enough.

PR #386, opened 2026-04-08
Add support to add/open cbz/cb7 files

apiweb:
Adds support for importing comic book archives (cbz/zip and cb7/7z).
The archives will get extracted on import, images are pulled out, sorted (I'm using `natord` to sort `page2` comes before `page10`), and treated as regular pages. 
I tried to change the minimum as possible, everything goes through the existing import pipeline.
Works for Open File and Add File.

I added some validations to avoid archive files with password or invalid or empty for example.

**Tests**
Tested with a ~1GB cbz (750 pages) file and it worked fine and fast.

I also added unit tests and e2e tests (for e2e I created a `cbz` and a `cb7` files using the existent fixtures pages). All green for the test suite.

**Demo**

https://github.com/user-attachments/assets/3194e0e8-1bb9-45f5-862e-f34f8008219d


mayocream (commented):
Please hold this; I will review it after redesigning the data model. 🙇‍♀️

PR #393, opened 2026-04-08
Adds the ability to reorder pages

apiweb:
Adds the ability to reorder pages through a `Page Manager` dialog with drag and drop support.

On frontend:
I added a new `PageManagerDialog` modal with a sortable grid of page thumbnails using @dnd-kit,
the navigator header shows a grid button when there are more than 2 pages to open the dialog.
Save persists the new order via the reorder API, Cancel discards changes.

On backend:
We have a new `PUT /api/v1/documents/order` endpoint that accepts an ordered list of page ids.
Update the `list_documents()` to sorts by `(order, name, id)` instead of just `(name, id)`.

**Tests**
I also added unit tests and e2e tests (testing four scenarios). All green for the test suite.

**Demo**

https://github.com/user-attachments/assets/cd6b7202-a96c-4b09-b66c-9f93a6548834


koharu-app/src/storage.rs
```diff
@@ -237,6 +237,36 @@ impl Storage {
         self.persist(&project)
     }
 
+    /// Reorder pages to match the given id sequence.
+    pub async fn reorder_pages(&self, ids: &[String]) -> Result<()> {
```
liksunrice, line 242: There is a bug due to how this function assigns `order = 0` that conflicts with `assign_missing_orders` will send an image that was sorted to that order be considered as 'missing order' and also gets moved to the end after adding a new image.

Consider reserving `0` as 'missing order' and apply a +1 offset to assigned orders using this logic.

https://github.com/user-attachments/assets/c26fdd29-fabb-468f-9bf4-3b4839b87884

liksunrice (changes requested):

koharu-app/src/storage.rs
apiweb: Great, I got it
My bad, I've fixed it, i completely lost that detail. I also added a unit test and e2e test

https://github.com/user-attachments/assets/40a6f98a-da62-4e13-a2e0-4ff269f96345

liksunrice (approved):
lgtm now!

PR #396, opened 2026-04-09
feat: Tool keybind and Brush Outline on canvas implementation

liksunrice:
## Summary
Implemented a simple brush outline that scales according to brush px size selection and zoom on image. This outline does not show if the appropriate mode selections are no selected, and only while the cursor is within the area of the canvas.

https://github.com/user-attachments/assets/e3e6807c-3f35-49b8-8fd1-ea7e136a01a8


karrot0:
Maybe we can add [ ] keybind to change brush size. Or some other keys for tablet / pen support

mayocream (approved):

PR #398, opened 2026-04-09
Add copy/paste formatting

apiweb:
I added a copy/paste formatting functionality to the Render controls panel, allowing users to copy text style from one block and paste it to another block. 

You can use the copy/paste buttons or the hotkeys Ctrl+Alt+C/Ctrl+Alt+V (on Mac: ⌘⌥C/⌘⌥V).

**Tests**
I also added new e2e tests covering the buttons, hotkeys, copying/paste styles, etc...

**Demo**

https://github.com/user-attachments/assets/64daae2b-e27c-4dcb-ba74-4a45ab4a3a35


karrot0 (dismissed):

karrot0 (dismissed):

karrot0 (approved):

mayocream (changes requested):
please hold

mayocream:
Copy/paste formatting is an incorrect design. I don't think modern design/editing software has copy-paste formatting; we should use multi-selection instead of this hack.

PR #418, opened 2026-04-13
docs: correct download button label

kidonng:
Current download button label is static: it states "Download for Windows" regardless of visitor's OS (despite releases for other OS are available); it also lists a fixed version (0.42.1) instead of the actual latest version (as of now 0.43.2).

There are two ways to address this issue:

A. Detect OS and fetch latest release
B. Simply don't show these info

Option B seems more practical, hence this pull request.

### DIFF
--- a/docs/en-US/index.md
+++ b/docs/en-US/index.md
@@ -111,11 +111,6 @@ hide:
     box-shadow: none;
   }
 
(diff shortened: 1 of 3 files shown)


mayocream (approved):
Thank you!

PR #480, opened 2026-04-16
ui: display language string clearly

fffonion:


ui/public/locales/en-US/translation.json
```diff
@@ -122,46 +122,46 @@
...
-      "ta-IN": "Tamil (India)",
-      "uk-UA": "Ukrainian (Ukraine)",
-      "bo-CN": "Tibetan (China)",
-      "kk-KZ": "Kazakh (Kazakhstan)",
-      "mn-MN": "Mongolian (Mongolia)",
-      "ug-CN": "Uyghur (China)",
-      "yue-HK": "Cantonese (Hong Kong SAR China)",
-      "pt-BR": "Brazilian Portuguese"
+      "en-US": "English",
+      "zh-CN": "中文（简体） | Simplified Chinese",
```
mayocream: I think the language list should be displayed in the same language of the language of UI. For example, google translate and apple time widget have the same logic.

mayocream: The language list of UI should be in native language instead, that’s how we proceed it nowadays.

ui/public/locales/en-US/translation.json
```diff
@@ -122,46 +122,46 @@
...
+      "zh-TW": "中文（繁體） | Traditional Chinese",
```
mayocream: "zh-TW": "Traditional Chinese",

Let’s remove the country name from the original language list, only kept if they are variants of a language.

josebrunoleite:
In line 292, I suggest changing it `"localLlmBaseUrl": "Base URL"` to `"localLlmBaseUrl": "URL base"`, as this is the correct wording in Brazilian Portuguese.

mayocream:
Fixed in https://github.com/mayocream/koharu/commit/56e078405483192753a5f025406d4f7bf2120199

PR #475, opened 2026-04-16
fix: RTL Support and Text Alignment Overhaul

liksunrice:
## Summary
Reworked the Right-to-Left (RTL) support to fix the disconnected Arabic letters issue detailed in #431 and overhauled the text alignment system or be more standardized throughout the application. This ensures that Arabic and Hebrew text renders correctly and flows well within speech bubbles alongside Latin texts.

## Changes
- Updated `detect_scripts` and `shaping_direction_for_text` to correctly identify and flag RTL text for the HarfBuzz shaping engine.
- Integrated horizontal and vertical alignment directly into `TextLayout` pipeline.
- RTL scripts now align center instead of left.
- Removed alignment math from `koharu-app` and delegated all positioning responsibility to `koharu-renderer`

| Before Changes | After Changes |
| --- | --- |
| beforearabic | fixedarabic |

PR #485, opened 2026-04-16
fix: revert fallback of `text_align`

### LINKED ISSUE #372: [Bug] Vertical alignment inconsistency in Korean (Hangul) text rendering
Description
There is an issue where the vertical alignment of the text differs depending on the language.
While English text is correctly centered vertically within the text block, Korean (Hangul) text is not positioned in the center, appearing shifted (usually upwards).

Steps to Reproduce

Translate a manga page into English and check the vertical alignment in the bubbles.

Translate the same page into Korean.

Compare the vertical positioning of the text in the same speech bubbles.

Actual Result

English: Text is perfectly centered vertically within the block.

Korean: Text is not centered vertically, causing an unbalanced visual appearance.

Expected Result
Korean text should also be rendered with the same vertical centering logic as English.

Screenshots

Image 

(English - Correctly centered)

Image

 (Korean - Misaligned)

### LINKED ISSUE #462: Rendered CJK text should horizontally center too.

Tiphereth-A:
Restored the fallback behavior of `text_align`.

<details>
<summary>Outdated message</summary>

fix #372
fix #462

Removed the check for `is_latin_only` so that all horizontal text is centered.

Example images with text blocks shown (text blocks with decorative & copyright content removed):

text_blocks

|lang|before (v0.44.5)|after|
|-|-|-|
|ar|ar_before|ar_after|
|ja|ja_before|ja_after|
|ko|ko_before|ko_after|
|th|th_before|th_after|
|zh|zh_before|zh_after|

</details>


Tiphereth-A:
I just noticed that #475 seems to have fixed the same issues, but it removed the fallback of `text_align`. Should we keep it?

Tiphereth-A:
I think the linked issues can be closed

PR #502, opened 2026-04-18
feat: add sequential (one-by-one) translation mode for better LLM stability

NotWeyn:
Description
Summary The current translation workflow batches all detected OCR texts on a page and sends them to the LLM in a single, massive prompt. While this is highly efficient for large models, it often causes problems for smaller parameter models (like local LLMs), leading to skipped texts, incorrect formatting parsing, and increased hallucinations.

This PR introduces a new "Translate one by one" option under the "Engines" section in the settings. When enabled, instead of sending the texts in a batch, the backend iterates through each text block and makes independent LLM translation calls for each block.

Changes Made

Backend (Rust): Added a translate_one_by_one flag to the PipelineConfig struct. Updated the llm_generate logic in llm.rs to loop over doc.text_blocks and translate them individually when this setting is enabled.
Frontend (Tauri/React): Extended the TypeScript schema and implemented a new UI <Switch /> toggle at the bottom of the Engines pane in SettingsDialog.tsx.
i18n (Localization): Added the translateOneByOne and translateOneByOneDescription keys to all 9 supported locales.

Motivation

To dramatically improve translation quality and prevent parsing deviations when using smaller/local LLMs (e.g., Ollama).
To provide users with the flexibility to choose between speed (batching) and accuracy/stability (sequential) depending on the capability of the specific model they are using.
Impact When enabled (optional), the overall translation time for a page will increase since the model is called multiple times sequentially. However, the trade-off is a much higher accuracy rate and fewer skipped speech bubbles for low-end local models. It is set to false by default.

Testing

 Rust backend compiled successfully without errors/warnings (cargo check).
 TypeScript types and lint checks have passed (tsc).
 Verified that toggling the setting properly updates and persists the config state in the UI.
 Confirmed the backend llm_generate successfully isolates translation requests when the flag is true.


mayocream:
This seems like a workaround rather than a true solution; feeding the LLM line by line is inefficient and loses context.

PR #518, opened 2026-04-20
fix(i18n): show language names in their native locale

Map1en:
Show app language choices using each language's native name instead of the active UI locale

### DIFF
--- a/ui/components/canvas/CanvasToolbar.tsx
+++ b/ui/components/canvas/CanvasToolbar.tsx
@@ -33,7 +33,6 @@ import {
   useGetCurrentLlm,
 } from '@/lib/api/default/default'
 import type { LlmCatalog, LlmCatalogModel, LlmProviderCatalog, LlmTarget } from '@/lib/api/schemas'
-import { getLanguageDisplayName } from '@/lib/i18n'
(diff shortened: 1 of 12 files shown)


ui/lib/i18n.ts
```diff
@@ -27,7 +27,25 @@ export const resources = {
   'pt-BR': { translation: ptBR },
 } satisfies Resource
 
-export const supportedLanguages = Object.keys(resources)
+export type SupportedLanguage = keyof typeof resources
+
+export const supportedLanguages = Object.keys(resources) as SupportedLanguage[]
+
+export const languageDisplayNames = {
```
mayocream: I think you can just change `menu.languages` in i18n

ui/components/canvas/CanvasToolbar.tsx
```diff
@@ -400,7 +401,7 @@ function LlmStatusPopover() {
                       value={language}
                       data-testid={`llm-language-option-${index}`}
                     >
-                      {t(`llm.languages.${language}`)}
+                      {getLanguageDisplayName(language) ?? t(`llm.languages.${language}`)}
```
mayocream: this is not correct, llm languages list should always show in local langauge

mayocream (approved):

PR #520, opened 2026-04-20
feat: implement .env file-based fallback for API key storage when keyring is unavailable or fails

NotWeyn:
This PR addresses an issue where API keys (Gemini, OpenAI, etc.) failed to persist or be used on Linux systems when a system keyring service (e.g., GNOME Keyring, KWallet) was unavailable or when the --no-keyring flag was used.

On some systems, a keyring service may not be available out of the box, or the user's keyring setup may be broken (e.g. no KWallet, broken gnome-keyring, headless environments). Previously, if the keyring was unavailable, the operation would silently fail with no error or warning shown to the user.
This PR introduces a .env file-based fallback for API key storage. If the keyring cannot be read from or written to, the app will automatically fall back to a .env file located next to config.toml in the app data directory. API keys will be stored and read from this file transparently, with no user intervention required. On Unix systems, the file is created with 600 permissions (owner read/write only).


mayocream (changes requested):
I think it's not the correct way to fix it: desktop software shouldn't use a `.env` file.
We should either fix the env fallback in #499 or guide the user to install keyring on Linux.

NotWeyn:
aaaaaah  I don't know, I managed to connect to the API outside of Docker, but when I run it inside Docker, it just doesn't work at all. When the API key is saved, the indicator light doesn't turn green or red, the same code works fine outside of Docker.

mayocream:
0.47.5 fixes this issue by adding a filesystem-based custom keyring implementation. Please try it out.

PR #542, opened 2026-04-22
Add regression test for global and individual font changes

liksunrice:
## Summary:
Fixes a reported issue where a global font change also affects individual text boxes with a custom font change selected.

### DIFF
--- a/ui/components/panels/RenderControlsPanel.tsx
+++ b/ui/components/panels/RenderControlsPanel.tsx
@@ -158,7 +158,8 @@ export function RenderControlsPanel() {
     [
       ...sortedFonts,
       ...(appDefaultFont ? [fallbackFontFace(appDefaultFont)] : []),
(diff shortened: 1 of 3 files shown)


ui/components/panels/RenderControlsPanel.tsx
```diff
@@ -321,10 +320,7 @@ export function RenderControlsPanel() {
                 currentFontFamilyName ? { fontFamily: currentFontFamilyName } : undefined
               }
               onChange={(value) => {
-                if (selectedNode) {
-                  applyStyleToSelected({ fontFamilies: [value] })
-                  return
-                }
+                if (applyStyleToSelected({ fontFamilies: [value] })) return
```
mayocream: Can you explain what this change is for?
The original `if (selectedNode) {` was for the scenario that only applies the style to the selected node, the global font change will override every node.

liksunrice:
Bug was already previously fixed, I have now changed this PR into implementing regression tests for the fixed issue

mayocream (approved):
Thank you!

PR #557, opened 2026-04-24
Fix: Brush and Mask Layer not syncing on undo/redo

liksunrice:
## Summary
This fix simply just exports the `invalidateScene` trigger from `scene.ts` and hooks it to `useMaskDrawing.ts` and `useRenderBrushDrawing.ts` to ensure that their respective layers stays in sync when using the undo/redo functions.

### DIFF
--- a/koharu-rpc/src/routes/pages.rs
+++ b/koharu-rpc/src/routes/pages.rs
@@ -8,13 +8,17 @@
 //! All three do the same server-side dance: read bytes → `blobs.put_bytes`
 //! → emit an `Op` on the session history.
 
+use std::sync::Arc;
+use std::sync::atomic::AtomicBool;
+
 use axum::Json;
 use axum::body::Bytes;
-use axum::extract::{Multipart, Path, State};
+use axum::extract::{Multipart, Path, Query, State};
 use image::GenericImageView;
+use koharu_app::pipeline::{self, EngineCtx, PipelineRunOptions};
 use koharu_core::{
     BlobRef, ImageData, ImageRole, MaskRole, Node, NodeDataPatch, NodeId, NodeKind, Op, Page,
-    PageId, Scene, Transform,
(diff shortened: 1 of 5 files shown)


liksunrice:
## New Changes:
- `PUT /pages/{id}/masks/{role}` endpoint optionally accepts pipeline and region parameters, allowing for mask update and inpainting to be in a single `Op::Batch`
- Frontend now triggers both actions in one undo/redo request, previous implementation required two requests to undo/redo both actions.

ui/hooks/useMaskDrawing.ts
```diff
@@ -68,54 +72,74 @@ export function useMaskDrawing({
...
+
+          const params = new URLSearchParams({
+            pipeline: inpainter,
+            x: region.x.toString(),
+            y: region.y.toString(),
+            width: region.width.toString(),
+            height: region.height.toString(),
+          })
+
+          const res = await fetch(`/api/v1/pages/${page.id}/masks/segment?${params}`, {
```
mayocream, line 97: let's use orval-generated hooks; do not call the API via fetch directly.

koharu-rpc/src/routes/pages.rs
```diff
@@ -509,7 +527,64 @@ async fn put_mask(
...
+        // Atomic Batch: Mask Update + Pipeline Run
+        let mut ops = vec![mask_op.clone()];
+
+        // 1. Simulate the mask update in a cloned scene so the engine sees it.
+        let mut scene = session.scene_snapshot();
+        mask_op
+            .apply(&mut scene)
+            .map_err(|e| ApiError::internal(e.into()))?;
+
+        // 2. Prepare EngineCtx
```
mayocream, line 540: I feel like calling engnie run after the ops if wrong 🤔

mayocream (approved):

PR #565, opened 2026-04-24
fix(ui): open updater release-note links externally

### LINKED ISSUE #564: Updater release-note links navigate inside the app webview
When the updater dialog renders release notes, external markdown links open inside the Tauri webview.

This traps users on GitHub or another external page because the app window does not provide browser navigation controls.

Expected behavior:
External release-note links should open in the system browser.

Actual behavior:
The app webview navigates to the external URL.

Clicking on release notes from this UI
Image

Leads to this inside the webview
Image

dragoon4890:
## Summary

Fixes #564.

Updater release notes are rendered from markdown. External links were rendered as normal anchors, which caused the embedded Tauri webview to navigate to GitHub or other external pages.

This adds a custom `ReactMarkdown` anchor renderer for updater release notes. External `http(s)` links now call `preventDefault()` and open through the existing `openExternalUrl()` helper, so they open in the system browser instead of replacing the app view.

## Testing

- Ran `bun --cwd ui lint`
- Verified in Tauri dev that clicking an updater release-note link opens the system browser
- Verified the Koharu webview stays on the app instead of navigating to the external URL

### DIFF
--- a/ui/components/Updater.tsx
+++ b/ui/components/Updater.tsx
@@ -20,7 +20,7 @@ import { Dialog, DialogContent, DialogDescription, DialogTitle } from '@/compone
 import { Progress } from '@/components/ui/progress'
 import { ScrollArea } from '@/components/ui/scroll-area'
 import { Separator } from '@/components/ui/separator'
-import { isTauri } from '@/lib/backend'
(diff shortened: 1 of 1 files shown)


mayocream (approved):
Thank you for the fix 😊

PR #567, opened 2026-04-25
fix: implement universal locale-aware uppercase conversion using ICU

edsporfekt:
What changed:

Integrated icu_casemap and icu_locale (ICU 2.2) to support universal, locale-aware case conversion.
Refactored normalize_translation_for_layout in koharu-renderer to accept an optional language identifier.
Updated PageRenderOptions and the rendering pipeline in koharu-app to pass the target_language down to the renderer.
Fixed a long-standing issue where languages with special casing rules (like Turkish) were incorrectly uppercased using standard ASCII/Unicode rules.
User-visible behavior differences:

Manga translations will now respect language-specific casing rules. For example, in Turkish, "kimse" will be correctly uppercased to "KİMSE" (with a dotted 'İ') instead of the incorrect "KIMSE".
This ensures better typography and readability for localized manga content.
How you verified the change:

Added new unit tests in koharu-renderer/src/text/script.rs specifically for:
Standard Latin uppercasing (e.g., English).
Locale-specific cases (e.g., Turkish 'i' and 'ı').
Ensuring non-cased scripts (like CJK) are not affected.
Verified all tests pass locally using cargo test -p koharu-renderer.
Verified code formatting with cargo fmt --check.
AI Usage Disclosure:

This PR was developed with the assistance of Antigravity. All code changes and logic were thoroughly reviewed and tested locally before submission to ensure they meet Koharu's quality standards.


mayocream (approved):
Thank you!

PR #572, opened 2026-04-25
fix: implement universal locale-aware uppercase conversion for all no…

edsporfekt:
What changed:
Universal Script Support: Refactored the uppercase normalization logic to support all non-CJK scripts (Cyrillic, Greek, Armenian, etc.). Previously, the system only allowed normalization for "Latin-only" text.
Removed Restrictive Checks: Replaced the is_latin_only validation with a more inclusive !is_cjk_text check, ensuring that mixed or non-Latin translations are correctly processed.
ICU Integration: Leveraged icu_casemap and icu_locale (ICU 2.2) to implement industry-standard, locale-aware case conversion.
Enhanced CJK Detection: Updated CJK script detection to include Hangul (Korean), ensuring it correctly bypasses casing logic.
Turkish Casing Fix: Specifically addressed the "dotted/dotless I" issue in Turkish. By passing the target_language from the UI to the renderer, translations now follow correct linguistic rules (e.g., i → İ).
User-visible behavior differences:
Multi-language Support: Translations in languages like Russian, Bulgarian, or Greek will now be correctly converted to uppercase in speech bubbles.
Turkish Typographic Accuracy: Turkish readers will see correct casing (e.g., ışık → IŞIK and kimse → KİMSE), preserving the intended typography and readability.
Robustness: Fixed issues where mixed-character strings or certain symbols would cause the normalization code to skip the entire block.
How you verified the change:
Comprehensive Unit Tests: Added new tests in koharu-renderer/src/text/script.rs covering:
Turkish-specific cases: Verified tr and tr-TR locales.
Other Scripts: Verified Cyrillic (привет) and Greek (γειά σου) conversions.
CJK Safety: Confirmed Han and Hangul remain untouched.
CI Readiness: Verified all tests pass using cargo test -p koharu-renderer.
Performance: Ensured ICU data access is efficient by using compiled data features.
AI Usage Disclosure:
This PR was developed with the assistance of Antigravity. All logic changes and refactors were reviewed, verified, and tested locally to ensure they align with Koharu's architectural standards.

### DIFF
--- a/koharu-renderer/src/text/script.rs
+++ b/koharu-renderer/src/text/script.rs
@@ -224,7 +224,6 @@ fn is_cjk_text(text: &str) -> bool {
             IcuScript::Han
                 | IcuScript::Hiragana
                 | IcuScript::Katakana
-                | IcuScript::Hangul
                 | IcuScript::Bopomofo
         )
     })
--- a/koharu-renderer/src/text/script.rs
+++ b/koharu-renderer/src/text/script.rs
@@ -33,18 +33,9 @@ pub fn writing_mode_for_block(block: &RenderBlock) -> WritingMode {
     }
 }
 
-pub fn is_latin_only(text: &str) -> bool {
-    let script_map = CodePointMapData::<IcuScript>::new();
-    text.chars().all(|c| {
-        matches!(
-            script_map.get(c),
-            IcuScript::Latin | IcuScript::Common | IcuScript::Inherited
-        )
-    })
-}
 
 pub fn normalize_translation_for_layout(text: &str, language: Option<&str>) -> String {
-    if is_latin_only(text) {
+    if !is_cjk_text(text) {
         let mapper = CaseMapper::new();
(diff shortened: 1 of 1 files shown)


koharu-renderer/src/text/script.rs
```diff
@@ -230,7 +221,11 @@ fn is_cjk_text(text: &str) -> bool {
     text.chars().any(|c| {
         matches!(
             script_map.get(c),
-            IcuScript::Han | IcuScript::Hiragana | IcuScript::Katakana | IcuScript::Bopomofo
+            IcuScript::Han
+                | IcuScript::Hiragana
+                | IcuScript::Katakana
+                | IcuScript::Hangul
```
mayocream: let's remove IcuScript::Hangul from the list, since `is_cjk_text` will be used for layout direction, and Korean text should always use horizontal

mayocream (approved):

PR #598, opened 2026-04-27
Add manual text direction toggle to render controls

### LINKED ISSUE #597: [Feature proposal] Add manual text direction toggle to render controls
## Problem

Currently, text writing mode is decided automatically in `koharu-renderer/src/text/script.rs::writing_mode_for_block`.

As far as I understand, the current behavior is roughly:
1. Empty text -> horizontal
2. Non-CJK text -> horizontal, regardless of `source_direction` or bbox shape
3. For CJK text, use `block.source_direction` if available
4. For CJK text without source direction, fall back to bbox heuristic:
   `height > width` -> vertical, otherwise horizontal

This works in many cases, but when the automatic direction detection is wrong, there is currently no way to correct it from the UI.

## Failure case

In the example below, vertically written Japanese text is recognized/rendered as horizontal.

Vertical Japanese text rendered horizontally

Expected vertical Japanese text rendering

## Why this would be useful

For manga translation/editing, text direction is not always reliably recoverable from OCR, layout detection, or bounding-box shape alone.

A manual direction control would allow users to quickly fix cases where:

- vertical Japanese/CJK text is detected as horizontal
- horizontal text is detected as vertical
- the text block shape does not match the intended writing mode
- OCR/layout detection produces an incorrect `source_direction`

## Proposed solution

Add a small text direction toggle in the render controls panel.

The toggle would allow users to manually switch the selected text block between horizontal and vertical direction.

I have implemented a small UI change for this and will open a PR linked to this issue. The current panel looks like below:

Render controls panel with text direction toggle

## Open question

While implementing this, I noticed that non-CJK text is still rendered horizontally even when the direction is set to vertical. This matches the current renderer behavior: non-CJK text returns horizontal before `source_direction` is checked.

If maintainers think this manual direction control is useful, I think the intended non-CJK behavior should be clarified before the UI is finalized. Otherwise, some users may see a clickable direction button that changes document state but produces no visible rendering change.

I see two consistent options:

1. Treat the direction toggle as a general explicit override.
   - The toggle can be shown for all text blocks.
   - The renderer should respect vertical `sourceDirection` for Latin / non-CJK text as well.

2. Treat vertical direction as a CJK-only feature for now.
   - This matches the current renderer behavior.
   - The UI should hide, disable, or clearly explain the toggle for non-CJK text blocks.

The current implementation is an initial proposal focused on the CJK manga correction workflow. I’m happy to adjust or simplify it based on the maintainers’ preferred direction.

jln-tail:
## Summary

This PR adds a manual text direction toggle to the render controls panel.

It allows users to switch selected text blocks between horizontal and vertical direction when automatic writing-mode detection misclassifies a block. This is mainly intended for CJK manga/OCR cases, such as vertical Japanese text being detected or rendered as horizontal.

Closes #597

## Changes

- Add a text direction toggle to `RenderControlsPanel`
- Toggle selected text blocks between `horizontal` and `vertical`
- Write the selected direction to `text.sourceDirection` through the existing op pipeline
- Add i18n labels for the new render control
- Add tests for selected-block and global text direction updates

## Current behavior / open question

This PR does not change renderer behavior.

Non-CJK text is still rendered horizontally even when `sourceDirection` is set to `vertical`, because the renderer currently returns horizontal for non-CJK text before checking `source_direction`.

If maintainers think this feature is useful, I’m happy to adjust the UI based on the intended behavior: either make vertical direction a general override for all scripts, or keep it CJK-only and hide/disable/explain the control for unsupported text blocks.

## Screenshots

image

## Testing

- Added tests for toggling text direction on a selected text block
- Added tests for applying text direction globally when no text block is selected
- Manually confirmed that the selected text block updates and the page re-renders

### DIFF
--- a/ui/components/panels/RenderControlsPanel.tsx
+++ b/ui/components/panels/RenderControlsPanel.tsx
@@ -4,6 +4,8 @@ import {
   AlignCenterIcon,
   AlignLeftIcon,
   AlignRightIcon,
+  ArrowDownIcon,
+  ArrowRightIcon,
   BoldIcon,
   ItalicIcon,
   MinusIcon,
@@ -31,6 +33,7 @@ import type {
   FontPrediction,
   Op,
   TextAlign,
+  TextDirection,
   TextShaderEffect,
   TextStrokeStyle,
   TextStyle,
@@ -191,6 +194,15 @@ export function RenderControlsPanel() {
     firstNode?.data.style?.textAlign ??
     (selectedNode?.data.translation ? 'center' : 'left')
 
+  const effectiveTextDirection: TextDirection =
+    selectedNode?.data.sourceDirection ??
+    selectedNode?.data.renderedDirection ??
+    firstNode?.data.sourceDirection ??
+    firstNode?.data.renderedDirection ?? 'horizontal'
+
+  const DirectionIcon =
+    effectiveTextDirection === 'vertical' ? ArrowDownIcon : ArrowRightIcon
+
   // ---------------------------------------------------------------------------
   // Mutations
   // ---------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


mayocream:

PR #609, opened 2026-04-28
fix: Revert inpainting logic for AOT/Lama

liksunrice:
## Summary:
This PR reverts the inpainting logic used by the AOT/Lama models prior to version 0.48.0 with the introduction of the Klein inpainting model due to a regression in the quality of output results with the current logic. This change attempts to replicate the original `expand_mask_for_inpainting` in `mask.rs` in release version 0.47.8, including its original parameters. 

| Inpainting Model | v0.53.0 | PR logic |
| --- | --- | --- |
| AOT | BeforeAOT | afterAO |
| Lama | BeforeLama | afterLama |


mayocream (approved):
Thank you!

PR #650, opened 2026-05-03
Skip already-satisfied pipeline steps

Selene29:
## Summary

Skip a pipeline step when all artifacts produced by that step are already present on the page.

This makes rerunning a pipeline less destructive and faster when earlier stages have already produced their expected outputs.

## Changes

- Check each engine step's declared `produces` artifacts before running it.
- If every produced artifact is already ready on the page, credit the step as complete and continue.
- Keep the existing pipeline order and failure behavior unchanged.

## Testing

- cargo test -p koharu-app pipeline

### DIFF
--- a/koharu-app/src/pipeline/mod.rs
+++ b/koharu-app/src/pipeline/mod.rs
@@ -180,6 +180,17 @@ pub async fn run(
                 continue 'pages;
             }
 
+            // Skip this step if its produced artifacts are already satisfied.
+            {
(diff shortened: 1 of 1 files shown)


liksunrice (commented):
Rerunning previously satisfied pipeline steps is the intended effect of the "Process Image" option. The suggested implementation of having this be an option separate to the current process is definitely the right approach.

PR #681, opened 2026-05-10
feat: huggingface mirror support

peterrobot:
## Summary

Add optional Hugging Face mirror support fror model downloads.

## Changes

- Add `huggingface_endpoint` to HTTP config and config patch handling.
- Pass the configured endpoint into runtime download setup.
- Update Hugging Face download code to prefer the app-configured endpoint, then fall back to `HF_ENDPOINT` from env variable.
- Add Runtime setting UI input for the Hugging Face mirror enddpoint.

## Known issue

Detector may failed to download without an error message, which may be related to missing package from the mirror site. But heavy model (like OCR / LLM) are tested to work fine.

Failure log as below, with `HF_ENDPOINT=https://hf-mirror.com`. Retrieved to default endpoint solve the problem.

```text
pipeline step failed: failed to download config.json from mayocream/speech-bubble-segmentation: failed to fetch HF metadata for mayocream/speech-bubble-segmentation/config.json: Header content-range is missing engine=speech-bubble-segmentation page=019df371-c7b7-77b2-b238-0b50a370e036 step_index=0
```

### DIFF
--- a/koharu-app/bin/pipeline.rs
+++ b/koharu-app/bin/pipeline.rs
@@ -144,6 +144,7 @@ async fn run() -> Result<()> {
         connect_timeout_secs: cfg.http.connect_timeout.max(1),
         read_timeout_secs: cfg.http.read_timeout.max(1),
         max_retries: cfg.http.max_retries,
+        huggingface_endpoint: cfg.http.huggingface_endpoint.clone(),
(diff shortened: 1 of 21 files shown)


mayocream:
we will never support this.

PR #736, opened 2026-05-27
feat: add per-block OCR action

GarryMS:
## Summary

Adds a per-block OCR action in the Text Blocks panel.

- Adds an OCR button next to the existing per-block translation action
- Sends `textNodeIds` when starting OCR from an individual block
- Updates PaddleOCR-VL, Manga OCR, and MIT 48px OCR engines to use `textNodeIds`
- Keeps existing full-page OCR behavior when no block scope is provided

## User-visible changes

Users can now run OCR for a single text block from the Text Blocks panel instead of rerunning OCR for every block on the page. Similar to how they can translate single text block. 

image

Existing page-level OCR behavior should be unchanged.

## AI Disclosure
Code Generation: I used Codex to draft the implementation.
- Translation: Also used Codex to help with Language translation (except English).
- Validation: I have built the project locally and test it manually.

### DIFF
--- a/koharu-app/src/pipeline/engines/manga_ocr.rs
+++ b/koharu-app/src/pipeline/engines/manga_ocr.rs
@@ -10,14 +10,16 @@ use koharu_ml::manga_ocr::MangaOcr;
 
 use crate::pipeline::artifacts::Artifact;
 use crate::pipeline::engine::{Engine, EngineCtx, EngineInfo};
-use crate::pipeline::engines::support::{load_source_image, text_node_to_region, text_nodes};
+use crate::pipeline::engines::support::{
+    load_source_image, scoped_text_nodes, text_node_to_region,
+};
 
 pub struct Model(MangaOcr);
 
 #[async_trait]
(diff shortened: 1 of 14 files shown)


mayocream (changes requested):
Thanks for this — per-block OCR still isn’t in Inspector/canvas, so the idea is useful.

This branch is about 463 commits behind `main`, and the files it patches no longer exist (`TextBlocksPanel.tsx`, the old `koharu-app` OCR engines, and the `ui/public/locales` copies). Current OCR already honors entity scope via `Scope::Entities` (`crates/koharu-pipeline/src/scope.rs` and `stages/ocr.rs`), so `textNodeIds` / `scoped_text_nodes` would be a parallel path.

Could you rework this on current `main`? The UI action should call `commands.process` with `{ scope: "entities", value: [layerId] }` and `{ operation: "only", stage: "ocr" }` instead of threading a new option through the old engines.

mayocream:

PR #738, opened 2026-05-28
feat: add the ability to import font from local file

GarryMS:
## Summary

Adds support for importing local font files into Koharu.

- Adds an import button beside the font controls
- Supports `.ttf`, `.otf`, `.ttc`, and `.otc`
- Copies imported fonts into Koharu’s app data directory
- Loads imported fonts into the renderer font book on startup
- Makes imported fonts available through the existing font selector

Imported fonts are treated as system fonts after they are loaded into the font book. I was trying to add a new FontSource but 1.I don't think defining a new source help with anything (except for use it to filtered font in the font select screen I guess?) and 2. it was a hassle since it need some changes to the API call

The scene continues to store the selected font by its normal PostScript name in `TextStyle.fontFamilies`, same as other fonts. Imported font files are only needed so the renderer can resolve that PostScript name later.

Font byte caching now stores all loaded face IDs for a font source instead of only one ID. This keeps duplicate imports consistent for font collection files like `.ttc` and `.otc`, where one file can contain multiple faces.

## Testing

I have built the project locally and tested the feature out manually.

## AI Usage

I used Codex to draft the implementation and help with language translations.

### DIFF
--- a/koharu-app/src/renderer.rs
+++ b/koharu-app/src/renderer.rs
@@ -19,6 +19,7 @@ use koharu_core::{
     TextStrokeStyle, TextStyle, Transform,
 };
 
+use camino::{Utf8Path, Utf8PathBuf};
 use koharu_renderer::{
     TextAlign as RendererTextAlign, TextShaderEffect as RendererEffect,
     font::{FaceInfo, Font, FontBook},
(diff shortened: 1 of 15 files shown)


mayocream:
Why is this PR required while fontdb already read system-wide installed fonts?

GarryMS:
> Why is this PR required while fontdb already read system-wide installed fonts?

Because I want to import and use some font that isn't on my system? For example Wild Words. I want to download it online and import straight to Koharu to use it.

I guess I can also download Wild Words and try to install it into my system, then Koharu can detect it but that's less intuitive, less portable and also more annoying

mayocream:
> > Why is this PR required while fontdb already read system-wide installed fonts?
> 
> 
> 
> Because I want to import and use some font that isn't on my system? For example Wild Words. I want to download it online and import straight to Koharu to use it.
> 
> 
> 
> I guess I can also download Wild Words and try to install it into my system, then Koharu can detect it but that's less intuitive, less portable and also more annoying

This isn't the correct way to handle fonts, we will require fonts to be installed on the system.

PR #742, opened 2026-05-29
[codex] Handle long-page slicing

salihozkara:
## Summary

- Add shared vertical slicing for very tall pages and use it across text detection, layout detection, speech-bubble segmentation, manga text segmentation, and Flux inpainting.
- Stitch sliced probability maps/masks with max blending, offset/dedupe detected regions, and keep inpainting scoped to masked slices before compositing back.
- Fall back from WebP to PNG for oversized image dimensions so rendered/inpainted outputs can be stored for long strips.
- Fix embedded asset lookup so the headless Web UI can serve bundled assets consistently.

## User-visible behavior changes

- Very tall manga strip images should no longer fail or over-resize as often during detection, segmentation, inpainting, and final render storage.
- Long-page processing may run as multiple internal slice passes, but the page remains a single imported page and the resulting masks, detections, inpainted image, and rendered image are stitched back into the original coordinate space.
- Oversized rendered/inpainted blobs may be stored as PNG instead of WebP when WebP dimension limits would reject the image; loading and export paths still decode the stored image normally.

## Root Cause

Very tall manga strips were being passed through several engines as one oversized frame. Some models resized the whole page too aggressively, some inpainting paths formed huge crops, and WebP encoding cannot handle dimensions above its limit. That made detect/OCR-adjacent segmentation/inpaint/render flows fail or become impractically slow on long images.

## Verification

- Ran `cargo fmt -- --check`
- Ran `git diff --check`
- Ran `koharu-ml` unit tests in a Linux build container
- Ran oversized WebP fallback test for `koharu-app`
- Smoke-tested the headless Web UI at `/` and `/api/v1/config`

## AI assistance

AI assistance was used to prepare this draft patch. The PR is intentionally kept as a draft until the contributor reviews and understands the changes before submission for maintainer review.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,10 +37,17 @@ jobs:
       - uses: actions/checkout@v6
       - uses: oven-sh/setup-bun@v2
       - uses: Swatinem/rust-cache@v2
+      - name: Free disk space
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc
+          sudo apt-get clean
+          df -h
       - uses: Jimver/cuda-toolkit@master
         if: matrix.os == 'ubuntu-latest'
         with:
           cuda: '13.0.0'
+          linux-local-args: '["--toolkit"]'
           log-file-suffix: '${{ matrix.os }}.txt'
       - uses: Jimver/cuda-toolkit@master
         if: matrix.os == 'windows-latest'
--- a/koharu-app/src/ai.rs
+++ b/koharu-app/src/ai.rs
@@ -257,7 +257,7 @@ impl AiManager {
                     let generated = image::load_from_memory(&generated_bytes)
                         .with_context(|| "failed to decode Codex image result")?;
                     let (width, height) = image_dimensions(&generated);
-                    let blob = session.blobs.put_webp(&generated)?;
+                    let blob = session.blobs.put_image(&generated)?;
                     Ok::<_, anyhow::Error>((width, height, blob))
                 })?;
             tracing::info!(width, height, "decoded and stored Codex generated image");
--- a/koharu-app/src/blobs.rs
+++ b/koharu-app/src/blobs.rs
@@ -12,13 +12,14 @@ use std::num::NonZeroUsize;
 use std::path::{Path, PathBuf};
 
 use anyhow::{Context, Result};
-use image::{DynamicImage, RgbaImage};
+use image::{DynamicImage, ImageFormat, RgbaImage};
 use koharu_core::BlobRef;
 use lru::LruCache;
 use parking_lot::Mutex;
 
 const IMAGE_CACHE_CAPACITY: usize = 64;
 const RAW_MAGIC: &[u8; 4] = b"RGBA";
+const WEBP_MAX_DIMENSION: u32 = 16_383;
 
 /// Content-addressed blob store + decoded-image LRU.
 pub struct BlobStore {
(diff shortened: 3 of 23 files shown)


salihozkara:
A quick note for transparency: I do not know Rust well enough to fully review the Rust implementation myself. I relied on Codex for the Rust changes and validated the branch with formatting and tests, but I would appreciate a careful Rust-level review for correctness and maintainability.

salihozkara:
Addressed the review feedback in the latest push:

- added a minimum slice height and removed the silent `num_slices.max(1)` fallback
- extracted shared bbox duplicate helpers and moved probability-map stitching onto `ProbabilityMap::stitch_max`
- updated Flux sliced inpaint to avoid reprocessing overlap pixels, fall through when no slice is processed, and crop page-sized reference images per slice
- preserved per-slice `line_polygons`, sorted layout regions deterministically before assigning `order`, and reused `VerticalSlicer::is_tall` for the AoT tall-page heuristic
- renamed image blob storage from `put_webp` to `put_image` and guarded the Tauri asset `/` fallback

Verification:
- `cargo fmt -- --check`
- `git diff --check`
- Docker/Linux: `cargo test -p koharu-ml --lib`
- Docker/Linux: `cargo test -p koharu-app oversized_webp_falls_back_to_png`
- Docker/Linux: `cargo test -p koharu-ml flux2_klein --lib`

Native Windows `cargo test` is still blocked by the missing MSVC `link.exe` in this process environment, so the Rust tests were verified in the Linux container.

salihozkara:
Follow-up review items from the latest pass are also addressed:

- kept `comic_text_bubble_detector` on its previous `target_slice_ratio: 3.0` while other long-page engines use the shared default
- sorted layout regions/text blocks before dedupe so higher-confidence duplicates win
- skipped non-matching full-size Flux references for slice-sized inpaint calls; matching page references are cropped per slice
- documented `VerticalSlicer::slices`, optimized CTD gray-map stitching over raw buffers, clarified `put_image` format expectations, and documented the Tauri asset fallback
- reordered speech-bubble probability stitching before offsetting region/mask coordinates so the final result consistently uses full-page coordinates

Re-ran:
- `cargo fmt -- --check`
- `git diff --check`
- Docker/Linux: `cargo test -p koharu-ml --lib`
- Docker/Linux: `cargo test -p koharu-app oversized_webp_falls_back_to_png`

salihozkara:
Addressed the newest review pass in `49c6134a`:

- `stitch_gray_max` now returns `Result` and validates width / destination Y instead of silently clipping
- `ProbabilityMap::stitch_max` now errors when `dst_y` is outside the destination map
- Flux sliced inpaint now crops from the running `output` so later slices see prior overlap context
- `VerticalSlicer::slices` documents that `MIN_SLICE_HEIGHT` can override the configured ratio for narrow images
- asset fallback now uses `prefixed` instead of shadowing `path`
- CTD `line_polygons` are deduped across overlapping slices

For the two optional items: I left O(n²) dedupe as-is because the region counts here are expected to stay small and the current code is simpler; I also kept oversized blobs as PNG fallback rather than adding tiled storage in this PR because tiling would change the blob/storage contract. I verified `LayoutRegion` only has `bbox` and `polygon_points` as positional fields, and both are offset.

Verification:
- `cargo fmt -- --check`
- `git diff --check`
- Docker/Linux: `cargo test -p koharu-ml --lib`
- Docker/Linux: `cargo check -p koharu`

Ex-Communicad0:
@salihozkara i am wondering what made you close the pr, is it not suitable, i am interested in webtoon format also let m eknow

salihozkara:
> @salihozkara i am wondering what made you close the pr, is it not suitable, i am interested in webtoon format also let m eknow

It was decided on the Discord channel that the app should not perform this task. The person who will use it should divide it into sections themselves beforehand. @Ex-Communicad0

PR #766, opened 2026-06-09
Feature/translation context

RavenZhou07:
## What changed

- Added translation context support for LLM translation.
- Added backend pipeline options for collecting prior text blocks/pages as context, including optional previous translations and a max context character budget.
- Added provider-level support checks and contextual translation entry points.
- Exposed translation context controls in the UI LLM settings popover:
  - enable/disable translation context
  - previous block count
  - previous page count
  - include previous translations
  - max context characters
- Persisted the new UI preference and updated the generated API schema/OpenAPI request shape.
- Added i18n strings for the new UI controls.
- Added manual OCR fixtures for comparing translation behavior with and without context.

## User-visible behavior differences

Users can now enable translation context for LLM translation runs. When enabled, Koharu passes nearby previous text blocks, and optionally their existing translations, to the LLM so that ambiguous terms, pronouns, names, and story references can be translated more consistently across a page or within pages.

In manual testing, this changed the translation of some ambiguous words. I created a manual OCR fixture and ran translation with context enabled and disabled. Some differences could verify that the context is working. For example, `spring` was translated as the season (`春天` / `春`) without context, and correctly translated as the mechanical spring (`弹簧` / `ばね`) when context was enabled. (if needed, I could provide the test file.)

The feature is enabled by default in the UI preferences, but it is only sent when the selected pipeline includes the LLM translator. Existing non-LLM translation paths should continue using the previous behavior.

## Verification
- Ran backend targeted tests:

  ```bash
  cargo test -p koharu-app llm_translate

## AI assistance
- AI helped produce parts of this patch. I am responsible for the submitted changes and have reviewed the code to understand and validate the implementation before submitting.


RavenZhou07:
The failing Windows build appears unrelated to this PR. The job enables the Tauri Windows `cuda` feature and installs CUDA 13.0, then fails while compiling `candle-kernels` because nvcc rejects the MSVC version on the current Windows runner:

unsupported Microsoft Visual Studio version
Only the versions between 2019 and 2022 are supported

Last night, my branch passed all 9 checks. After main changed today, I merged main from the PR page and saw the new checks fail. I then reverted my branch back to the last version that had passed those previous checks, but the updated checks are still running against the PR and the Windows build now fails with the CUDA/MSVC toolchain error above.

So this looks like a new CI/toolchain issue introduced by the updated checks rather than a regression from this PR’s application-level changes. Once the main branch CI/toolchain issue is fixed, I’ll merge the latest main changes again from the PR page.

mayocream (changes requested):
Thanks for the fixtures and the continuity idea — the spring vs season example is a nice demonstration.

This can't land as-is, though:

**Stale vs main.** The branch still targets the old `koharu-app` / `koharu-llm` / `ui` layout and currently conflicts with `main`. Translation now lives in `crates/koharu-pipeline` + `crates/koharu-translator`, with UI under `packages/koharu`.

**Parallel context path.** Main already has `TranslationRequest.context` (`TranslationContext { source, translation }`) wired through the JSON prompt. The pipeline just doesn't fill it yet. Please collect previous blocks/pages into that existing field (and if knobs are needed, hang them on `TranslationConfig`) rather than a new `TranslationContextConfig` on the run request, a `translate_texts_with_contexts` entry point, or per-provider `supports_translation_context` flags.

That also keeps this aligned with #1007, which already owns the `TranslationConfig` / system-prompt path. A second prompt format would fight that.

**Overlap with #830.** Open PR #830 (chapter context translation mode) is another second way to pass context into translation. Let's not ship two parallel context features — pick one approach on the current translator path.

Happy to re-review once this is rebased onto current main using `TranslationRequest.context`.

mayocream:

PR #801, opened 2026-06-19
fix: Various Delete Function Fixes

liksunrice:
## Summary
This PR fixes various bugs related to the delete function:
1. Selecting a page then selecting a text block on the canvas and pressing the delete key unintentionally deleted the entire page due to the document focus remaining on the navigator page preview card.
2. Refactored multiple text block deletions to execute under a singular `Op::Batch`.
3. `Op::Batch` also applied to batch the automatic re-rendering pipeline that automatically occurs after deleting a text block.

## Changes:
- Made `TextBlockItem` focusable upon selectable on click.
- Refactored `removeSelected` to handle the batching of multi-block deletions.
- Refactored `push_undo(&mut self, op:Op)` to automatically group auto-render pipeline runs with previous actions when applicable.


mayocream (approved):
LGTM!

PR #808, opened 2026-06-22
feat: added a "delete selected textblocks" button and keyboard shortcut

obsgolem:
This adds a few small tweaks to the textblock workflow. First, there is now a delete all button. Second, there is a deselect all button in the menu. Third, the delete key is mapped to delete all selected.

The delete button is put in a panel which is hidden until multiple selections are made. In the future, this panel can be used for other selection-specific buttons.

Note I didn't add a keyboard shortcut for clear selection. That shortcut should probably be escape, but I wasn't sure if it was desirable.

No AI assistance was used to write this MR.


obsgolem:
I just noticed that there are several other delete hotkeys. Gonna refactor them a bit so they all use useHotkey and handle focus correctly.

mayocream (approved):
LGTM

Issue #555, opened 2026-04-23
[Feature Request] Global Structured OCR Export (XML/HTML) with Speaker Assignment for LLM Integration


ChenNdG:
## Is your feature request related to a problem? 
Currently, Koharu is an excellent tool for OCR and text editing on a per-page basis. However, when using Large Language Models (LLMs) for translation, the AI often lacks the global context of the project (story arc, character gender, tone consistency). Manually copying text from each bubble is time-consuming and loses the structural relationship between pages and speakers.

## Describe the solution you'd like
I suggest implementing two main features to bridge the gap between OCR and AI-assisted translation:

### 1. Speaker Identification in the UI
* Add a metadata field in the **Text Blocks** panel to assign a speaker/character to each OCR block.
* A project-wide character list to quickly select names via a dropdown for consistency.

### 2. Global Structured Export
* A feature to export the entire project’s text into a single structured file (XML or HTML format).
* The export should preserve the hierarchy: **Project > Page > Block (with Speaker ID) > Text**.

**Example of the desired output (XML):**
```xml
<project name="Project_Title">
  <page id="1">
    <block id="1" speaker="Narrator">
      いつものようにゴミ出しをしていると…
    </block>
    <block id="2" speaker="Woman_A">
      よいしょ…と、…んっ？なんだろ？
    </block>
  </page>
</project>
```
## Describe alternatives you've considered
Currently, the only way is to copy-paste blocks one by one, which prevents the LLM from "seeing" the whole story at once, leading to translation errors regarding pronouns, tone, and character-specific speech patterns.

## Additional context
* **LLM Context Window:** Modern LLMs can handle entire scripts. Providing a full XML/HTML file allows the AI to understand the relationship between characters before translating.
* **Workflow Efficiency:** This would allow users to pre-process a whole chapter and then re-import the translated text more efficiently.
* **UI Integration:** 

<img width="900" height="725" alt="Image" src="https://github.com/user-attachments/assets/5a6e62b0-48bf-42c7-8f30-688aa5edb60d" />

PR #829, opened 2026-06-28
feat: Add JSON import / export of source and translation text

spaghetti-stack:
## Changes
- Allows exporting source and translation text of a whole project to JSON format. 
- Allows importing translation text for a whole project in JSON format.

This way it is now possible to use an external translation provider to, for example:
- Perform translations using an LLM that make use of the full context of all pages.
- Use external translators not currently supported by Koharu. 
- Compare and switch between multiple translations without having to rerun the translation pipeline.

This feature is similar to #614 and #555 (but without speaker ID).

I chose JSON as format because it is easy to parse, and the format is (slightly) more compact than XML so using an LLM for whole context translation would use less tokens.  XML or Markdown could also be added in the future.

## User facing changes:
image

Added the "Export source texts", "Export translations" and "Import translations" buttons to the UI which are used to interact with this new feature.

## Testing
I have manually tested the import / export functionality. There are also front end and back end unit tests.


mayocream (changes requested):
Thanks for the first PR — whole-project JSON import/export is a useful workflow for external translators / full-context LLM work.

1. **Rebase onto current main.** This branch is ~413 commits behind and the old HTTP RPC stack is gone. Export is now a Tauri+specta command in `crates/koharu-app/src/commands/output.rs` (`export_pages`, PNG/PSD only), the File menu lives in `packages/koharu/components/app/TitleBar.tsx`, and TS types are generated `packages/bridge/src/protocol.ts` (`// generated by koharu; do not edit`). Please add a command and let specta regenerate the protocol — don’t hand-edit OpenAPI/orval clients.

2. **Overlaps with #880.** That PR also added whole-project source/translation import-export (tagged `.txt` reusing the LLM `[n]` format). We shouldn’t land both. JSON is the approach I’d keep if we pick one — please rebase this onto current main rather than adding a second path.

3. **Apply path.** The batch-via-history idea is right. When rewriting, apply translations through the existing `set_translation` commit path so undo works.

mayocream:

PR #845, opened 2026-07-03
fix: fix bug where progress indicator doesn't progress

obsgolem:
The progress indicator for me frequently fails to update. Adding a manual yield to the scheduler appears to fix this.

No AI was used in the production of this MR.


mayocream (dismissed):
Thank you for catching this bug!

mayocream (approved):

PR #849, opened 2026-07-07
chore(dev): Switched to Microsoft recommended way of finding MSVC

obsgolem:
`vswhere` is the standard tool for finding VS installations. Switch to using that. This also allows us to support VS Insiders.

No AI was used in the creation of this MR.


mayocream (approved):

mayocream:
Thank you!

PR #851, opened 2026-07-08
fix: Allow deleting font size value completely.

obsgolem:
When typing a font size, an intuitive way to do that is to delete the whole thing and type in the number you want. This allows you to do that, and has the convenient side effect of allowing you to reset the font size to auto.

Note: it would be ideal to do this with border stroke width too, but we don't have an empty default state for that field. A future MR should add a default state and similar logic to that field.

This patch was made with no AI assistance.


obsgolem:
Hmm, there appear to be unrelated style changes. Must not have vscode set up right.

obsgolem:
> Hmm, there appear to be unrelated style changes. Must not have vscode set up right.

Resolved

mayocream (approved):
LGTM

PR #854, opened 2026-07-09
chore(ui): cleaned up the styling of the text style buttons

obsgolem:
Changed the icon when unselected to make it clear it is unselected. Adjusted colors so that in dark mode it is clearer that the button is selected and not deselected.

Currently I adjust the weight of the `SquareIcon` so that it looks less ugly in light mode. Ideally we would do this for all the icons when selected, and only in light mode. This is because the accent-foreground looks bad against accent at low font weights/stroke widths. Alternatively, we might play with the colors a bit more.

Note: the dark colors are a little bit ad-hoc right now. It might be worth reviewing the color scheme as a whole and creating a larger set of color tokens that better matches the semantic roles we have in play.

No AI was used in making this MR.

### DIFF
--- a/ui/app/globals.css
+++ b/ui/app/globals.css
@@ -145,3 +145,4 @@
     @apply select-text;
   }
 }
+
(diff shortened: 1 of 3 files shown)


mayocream (approved):

mayocream:
It would be helpful if you could include the UI screenshots in the PR description next time.

PR #856, opened 2026-07-10
Merge auto-render ops with the op just before it

### LINKED ISSUE #852: Multiple undo actions required to undo some operations
Sometimes I need to press ctrl-z twice to get the undo to actually happen.

For example: edit font size from "auto" to "8". Then press ctrl-z. Nothing happens. Press ctrl-z again, font size changes to "auto".

Maybe the render is generating an extra undo action?

obsgolem:
This ensures that auto-render is considered as part of a batch edit, thus preventing situations where ctrl-z appears to have no effect.

Note that I don't update the record on-disk, so when reloading it will be treated as a separate entry. Let me know if you want me to try to figure out a way to resolve this as part of the MR.

Fixes: #852

No AI was used in this MR.


mayocream:
This would work, but I wonder if we can "batch" auto-render with the previous op. 🤔

obsgolem:
> This would work, but I wonder if we can "batch" auto-render with the previous op. 🤔

The problem is the debouncing. The previous op needs to apply before the ui can update, so we can't afford to wait a debounce cycle. Thus my idea to merge it later and pretend like it was a batch in the first place.

liksunrice:
I think #801 fixed it via batching with previous op

mayocream (changes requested):
Thanks for this — merging auto-render with the preceding edit is the right way to stop Ctrl+Z from looking like a no-op.

This can't land as-is though: every file here is gone from `main` (scene-graph refactor). `history.rs` / `session.rs` / the old pipeline + RPC start-pipeline types / `ui/` no longer exist. Undo is now `Project::{record,record_commit}` over scene revisions, and canvas render is a derived `Desktop` presentation, not an undoable pipeline op.

#801 already batched renderer ops with the previous user edit in the old `History::push_undo` (as already noted here). Could you rebase onto main and check whether #852 still reproduces? If a font-size change still takes two Ctrl+Z, the port belongs next to `record_commit` / pipeline commits — not a new `apply_merge_up`. If one undo already reverts the edit, we can close this.

mayocream:

PR #860, opened 2026-07-11
Fix scaling on sprites

obsgolem:
For some reason, the browser decides to downscale the sprite's bounding box before even applying the transform. This causes the scale and translate values to be incorrect. Fix this by forcing width and height to the correct values.

Correct Render:
image

Preview Before:
image
Preview After:
image

### DIFF
--- a/ui/components/canvas/TextBlockLayer.tsx
+++ b/ui/components/canvas/TextBlockLayer.tsx
@@ -251,6 +251,8 @@ function BlockSprite({ node, scale }: { node: TextNodeEntry; scale: number }) {
   const spriteT = node.data.spriteTransform
   const x = (spriteT?.x ?? node.transform.x) * scale
   const y = (spriteT?.y ?? node.transform.y) * scale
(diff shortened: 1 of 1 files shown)


mayocream (changes requested):
Thanks for tracking this down — forcing explicit width/height instead of `transform: scale()` is the right way to stop the browser from shrinking the sprite box before the transform.

This can't land as-is though: `ui/components/canvas/TextBlockLayer.tsx` (and the whole `ui/` tree) is gone from `main` (scene-graph refactor). Overlay frames now live in `packages/koharu/components/editor/CanvasOverlay.tsx`, and they already size via `cssFrame` width/height rather than CSS `scale()`. `BlockSprite` / `spriteTransform` aren't on main anymore.

Could you rebase onto main and check whether the preview-vs-render mismatch still happens? If it does, it would be in the GPU canvas path (`crates/koharu-canvas`), not a straight copy of this img style fix.

Same situation as #861.

mayocream:

PR #729, opened 2026-05-26
fix: update color when picker drag ends outside the selection box

MOULINIER:
(First commit to a major project, sorry if I made any mistakes on the PR) 

**Summary**
When picking a color on the color picker box, it bugged me that when selecting colors at the corners like #FFFFFF or #000000 they weren't actually selected. It seems that when the cursor ends up outside of the box when selecting, the color isn't registered.

**Changes**
- Use react-colorful built-in onChangeEnd (added recently in 5.7.0) to mitigate the issue, as the feature was made for issues like this (https://github.com/omgovich/react-colorful/issues/201)
- Removed the handlePointerUp function / useCallback import

**Tests and AI**
I ran the integrated bun run --filter ui test.
Tested the fix manually.

I used AI to help me debug and learn the structure (I am a CS student and still learning!)

PR #861, opened 2026-07-11
Fix bug where color picker fails to update color

obsgolem:
This occurs when the mouse moves out of bounds of the color picker widget. This is due to the pointerUp event only firing inside the div. Fix this by using html pointer capture api.

### DIFF
--- a/ui/components/ui/color-picker.tsx
+++ b/ui/components/ui/color-picker.tsx
@@ -55,6 +55,10 @@ export function ColorPicker({
     if (!dragging.current) setLocalColor(value)
   }, [value])
 
+  const handlePointerDown = (event: React.PointerEvent) => {
+    event.currentTarget.setPointerCapture(event.pointerId)
+  }
+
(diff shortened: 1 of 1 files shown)


MOULINIER:
Already did in https://github.com/mayocream/koharu/pull/729 lol

mayocream (changes requested):
Thanks for the pointer-capture fix — that's the right way to keep `pointerup` firing when the pointer leaves the picker.

This can't land as-is though: `ui/components/ui/color-picker.tsx` is gone from `main` (scene-graph refactor). Color picking now lives in `packages/koharu/components/controls/ColorWell.tsx`. Could you rebase onto main and port the capture there? ColorWell doesn't use the old wrapper `onPointerUp` commit path, so it may need a small adaptation rather than a straight copy.

#860 / #856 don't overlap with this. #989 (ColorWell history) is already approved and shouldn't collide with a small pointer-capture change.

mayocream:

PR #868, opened 2026-07-21
fix: stop RAM growth during full pipeline runs

tqbdev:
## Problem

Running the pipeline over **all pages** (detect → OCR → translate → inpaint → render) made RAM climb monotonically page after page and never release it, building to large consumption by the end of a batch. Reproduced on macOS with the **LaMa inpainter on Metal**.

## Root causes

Three independent leaks (a fix for any one alone barely moved the needle):

1. **Frontend — object URLs never revoked.** `useBlobImage` (`ui/hooks/useBlobData.ts`) created `URL.createObjectURL(blob)` but never revoked it, so every sprite/inpaint/render blob pinned a full decoded image in the webview for the whole run. Engine-independent, so backend changes didn't help it.
2. **Backend — no Objective-C autorelease pool on Metal.** candle's Metal ops and the MPSGraph FFT (`lama/fft/metal.rs`) allocate autoreleased objects (command buffers, MPS intermediates); a leaked command buffer also pins every GPU buffer it referenced. The pipeline runs on tokio worker threads that have no autorelease pool draining, so these accumulate for the entire run. There was no `autoreleasepool` anywhere in the tree.
3. **Backend — unbounded ML caches.** The LaMa FFT-plan caches (keyed by crop shape, `lama/fft/{metal,cuda}.rs`) and the flux2 Qwen prompt cache (`flux2_klein/qwen.rs`) lived on whole-run engine instances and were never evicted.

## Fixes

- `ui/lib/queryClient.ts`: revoke a `blobImage` query's object URL when it's evicted from the query cache; `ui/hooks/useBlobData.ts`: drop blob `gcTime` 10min → 1min so inactive pages release promptly.
- `lama/mod.rs` + `lama/fft/metal.rs`: wrap the LaMa per-crop forward and the FFT calls in `objc2::rc::autoreleasepool` (no-op off Metal via a small cfg-split helper).
- `lama/fft/{metal,cuda}.rs` and `flux2_klein/qwen.rs`: bound the caches with `lru::LruCache`.

## Verification

- `cargo check -p koharu-ml` passes **with and without** the `metal` feature.
- `cargo test -p koharu-ml --features metal --lib lama::` passes.
- UI lint clean (no new warnings).
- Manually confirmed on macOS/Metal: RAM now plateaus during "process all pages" instead of climbing to the end.

## Notes

- If growth ever shows up specifically during Detect/OCR (not Inpaint), it's the same candle-Metal autorelease issue in those engines and can be fixed with the same wrap.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4354,6 +4354,7 @@ dependencies = [
  "image",
  "imageproc",
  "koharu-runtime",
+ "lru",
  "objc2",
  "objc2-foundation",
  "objc2-metal",
--- a/crates/koharu-app/src/pipeline/engines/anime_text.rs
+++ b/crates/koharu-app/src/pipeline/engines/anime_text.rs
@@ -20,7 +20,7 @@ pub struct Model(AnimeTextDetector);
 impl Engine for Model {
     async fn run(&self, ctx: EngineCtx<'_>) -> Result<Vec<Op>> {
         let image = load_source_image(ctx.scene, ctx.page, ctx.blobs)?;
-        let det = self.0.inference(&image)?;
+        let det = koharu_ml::autorelease_scope(|| self.0.inference(&image))?;
 
         let mut pairs: Vec<([f32; 4], TextData)> = det
             .text_blocks
--- a/crates/koharu-app/src/pipeline/engines/aot.rs
+++ b/crates/koharu-app/src/pipeline/engines/aot.rs
@@ -58,7 +58,8 @@ impl Engine for Model {
             None => DynamicImage::ImageLuma8(expanded),
         };
 
-        let result = self.0.inference(&image, &mask, &bubble_mask)?;
+        let result =
+            koharu_ml::autorelease_scope(|| self.0.inference(&image, &mask, &bubble_mask))?;
         let (w, h) = image_dimensions(&result);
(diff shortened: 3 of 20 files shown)


mayocream (changes requested):
Thanks for the careful write-up — the three leaks were real on the old candle/Metal + `useBlobImage` stack.

This branch is ~400 commits behind `main`, though, and those sites have already been rewritten:

- Frontend blob URLs no longer go through `useBlobImage` / `queryClient`. `PageRail` and `FontPicker` create object URLs and revoke them on effect cleanup.
- LaMa now runs on `koharu_torch`; `lama/fft/{metal,cuda}.rs` and the MPSGraph plan cache are gone.
- FLUX.2 Klein now runs on `koharu_diffusion`; `flux2_klein/qwen.rs` and the unbounded prompt cache are gone.
- Pipeline engines moved to `koharu-pipeline`. There is no `koharu_ml::autorelease_scope` on main (candle/objc2 is gone).

So this patch cannot land as-is. Could you rebase onto current `main` and re-measure a full pipeline run? If RAM still climbs, a new patch against the torch/diffusion + `packages/koharu` code is the way; if it already plateaus, we can close this.

mayocream:

PR #876, opened 2026-07-26
fix: unlock font-size field for global editing without selection

### LINKED ISSUE #649: Enable global font size override from "Auto" across all images
Problem Details
Currently, when the font size is set to "Auto", the font size input field becomes completely disabled (locked/unclickable).

Because of this:

It is impossible to override the "Auto" setting globally when no box is selected.

To change the font size of all boxes across multiple images, I have to manually go into each page, select all boxes (Ctrl + A), and then change the size.

The global font size setting should be the "Master" control, but it is currently unusable if the default state is "Auto".

Suggested Change

Please unlock (enable) the font size input field even when "Auto" is the current state in global settings.

Entering a value in the global font size field should automatically turn off "Auto" for all text boxes in the project and apply the specified size.

Why this is critical
In projects with hundreds of images, the inability to click and change the global font size from "Auto" forces hours of repetitive manual work. Making this field interactive would significantly improve the workflow for bulk typesetting.

JoeJoeflyn:
The font-size field was disabled when no text layer was selected, blocking the global workflow: typing a value with nothing selected should apply to every text layer on the page, and clearing it should reset all layers to auto.

**What changed:** Extracted all text layers into `textLayers` and changed the disabled guard from `!current` to `textLayers.length === 0`, so the field stays enabled whenever the page has text layers. In `apply`, fall back to `textLayers` when nothing is selected instead of returning early. Per-layer edits when a layer is selected are unchanged. `FontSizeField` already has its own draft buffer, so no separate global input state is needed.

**User-visible behavior differences:** With no text layer selected, the font-size field is now editable (previously greyed out). Typing a number applies it to every text layer on the page; clearing the field resets all layers to auto. Selecting a single text layer still applies changes only to that layer, unchanged.

**How I verified:** `bun run format` and `bun run test` (38/38 editor component tests passed, including a new test covering the no-selection case).

Fixes: #649

This patch was made with AI assistance (Devin/Claude). The code was reviewed and tested before submission.

### DIFF
--- a/ui/components/panels/RenderControlsPanel.tsx
+++ b/ui/components/panels/RenderControlsPanel.tsx
@@ -148,6 +148,14 @@ export function RenderControlsPanel() {
 
   const sectionRef = useRef<HTMLDivElement>(null)
   const [sectionWidth, setSectionWidth] = useState<number>(0)
+  // Buffers font-size input in global scope (no selection) so multi-digit
+  // values can be typed before committing — the controlled `value` would
+  // otherwise reset to '' on each keystroke because `currentFontSize` stays
+  // undefined when nothing is selected.
+  const [globalFontSizeInput, setGlobalFontSizeInput] = useState('')
+  useEffect(() => {
+    setGlobalFontSizeInput('')
+  }, [selectedNode?.id, page?.id])
 
   useEffect(() => {
     if (!sectionRef.current) return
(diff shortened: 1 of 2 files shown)


mayocream (changes requested):
Thanks for this — unlocking font-size with no selection is the right fix for #649, and `applyStyleToAll` vs `applyStyleToSelected` matches the old panel.

This branch is ~400 commits behind `main` though, and the files it edits are gone. Font size now lives in `packages/koharu/components/editor/Inspector.tsx` (`FontSizeField`), which is still disabled when nothing is selected (`disabled = !current`; `apply` returns if `!selected.length`). Could you rebase onto current `main` and apply the same idea there? Keep per-node edits when a layer is selected; apply to all text layers on the page otherwise. `FontSizeField` already has a draft buffer, so a separate `globalFontSizeInput` probably isn't needed.

Not a conflict with #914 — that's typesetting-preference auto/manual overrides, a different path.

mayocream:

JoeJoeflyn:
yo just woke up and got my pr closed, i will update it an push

PR #880, opened 2026-07-29
Add export and import options for original and translated scripts

LetrixZ:
Added an option to export the original script generated by the OCR and to import it as translated.

image

Single page example:
```
[1] Foo
[2] Bar
[3] Mutli
line
string
```

Multi page example:
```
Page 1
[1] Foo
[2] Bar
Page 2
[1] Mutli
line
string
```

This PR was AI assisted for the importing stuff and menu translations.


mayocream (changes requested):
Thanks for the first PR — exporting OCR text and importing it back as translation is a useful workflow for external translators / full-context LLM work.

1. **Rebase onto current main.** This branch is ~401 commits behind and the old HTTP RPC stack is gone. Export is now a Tauri+specta command in `crates/koharu-app/src/commands/output.rs` (`export_pages`, PNG/PSD only), the File menu lives in `packages/koharu/components/app/TitleBar.tsx`, and TS types are generated `packages/bridge/src/protocol.ts` (`// generated by koharu; do not edit`). Please add a command and let specta regenerate the protocol — don’t hand-edit OpenAPI/orval clients.

2. **Overlaps with #829.** That PR also adds whole-project source/translation import-export (JSON). This one adds a second path (tagged `.txt` reusing the LLM `[n]` format). We shouldn’t land both. Could you pick one approach and align with it rather than two ways to move scripts in/out?

3. **Round-trip / apply path.** A 1-page project currently omits the `Page N` header, so “Import all” cannot parse what “Export all” just wrote. When rewriting, apply translations through the existing `set_translation` commit path so undo works.

LetrixZ:
I don't have time to check the refactor in the main code.
Might re-do this in the future from scratch instead.

PR #881, opened 2026-07-30
feat(llm): add Atlas Cloud provider

binyangzhu000-sudo:
## Summary

- register Atlas Cloud as a first-class LLM provider
- reuse the existing OpenAI-compatible translation adapter with the managed `https://api.atlascloud.ai/v1` endpoint
- expose `deepseek-ai/deepseek-v4-pro` through the existing provider catalog and settings flow
- require an API key while keeping the base URL managed by the provider

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p koharu-llm --lib` (29 passed, 10 existing runtime-dependent tests ignored)
- `cargo check -p koharu-llm --lib`
- live Atlas Cloud translation through `build_provider("atlascloud", ...)` returned non-empty text
- `git diff --check`

## AI usage disclosure

AI tools assisted with implementation and test planning. I reviewed the provider flow and diff, ran the focused and full library checks above, and verified the actual Atlas Cloud runtime path.

### DIFF
--- a/crates/koharu-llm/src/providers/mod.rs
+++ b/crates/koharu-llm/src/providers/mod.rs
@@ -118,6 +118,13 @@ pub struct ProviderConfig {
     pub max_tokens: Option<u32>,
 }
 
+const ATLASCLOUD_BASE_URL: &str = "https://api.atlascloud.ai/v1";
+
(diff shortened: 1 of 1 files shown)


mayocream (approved):
LGTM

PR #900, opened 2026-08-12
fix(runtime): discover cached ROCm installations

louisoes05-png:
Hopefully makes all amd gpu workable

### DIFF
--- a/crates/koharu-runtime/src/hardware/hip.rs
+++ b/crates/koharu-runtime/src/hardware/hip.rs
@@ -1,6 +1,8 @@
-use std::ffi::c_void;
+use std::{ffi::c_void, path::PathBuf};
 
 use libloading::Library;
+
+use crate::Store;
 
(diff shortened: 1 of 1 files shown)


mayocream (changes requested):
Thanks for this — walking the cached ROCm store after the system HIP names was a reasonable idea on the old `amdhip64` probe.

`main` has moved on here, though: HIP discovery no longer loads the HIP runtime at all (KFD on Linux, OpenCL ICD on Windows), and it is intentionally kept apart from Koharu's managed ROCm package. That package is still installed and activated later in `runtime/packages/rocm.rs` once the gfx target is already known. Mixing `Store` into hardware probe would go the other way.

This branch also conflicts with `main` (`hip.rs` is now `hardware/hip/`). Could you rebase and check whether the original AMD GPUs still fail to show up? If they do, the fix belongs in the KFD/OpenCL probes rather than loading cached `amdhip64`.

PR #912, opened 2026-08-14
fix: recover from an out-of-range detection threshold

### LINKED ISSUE #911: Out-of-range detection threshold in config.toml prevents Koharu from starting
### Summary

A detection confidence threshold saved outside `0.0..=1.0` makes Koharu unopenable. Startup aborts before the window is usable, and the only way back in is to hand-edit `~/.koharu/config.toml`.

I hit this after typing `15` into **Preferences → Pipeline → Text threshold**. Every subsequent launch dies with:

```
Something went wrong and Koharu needs to close.

failed to initialize the desktop runtime: text confidence threshold must be
finite and between zero and one

at crates\koharu-app\src\app.rs:135
```

### Steps to reproduce

1. Open **Preferences → Pipeline**.
2. Type `15` into **Text threshold** (typing, not the stepper arrows).
3. Restart Koharu.

Or equivalently, put this in `~/.koharu/config.toml`:

```toml
[pipeline.processor.koharu-layout-rfdetr-seg-2xl]
text_threshold = 15.0
```

### Why it happens

Two independent problems line up:

1. **The value can be saved.** `NumberField` in `packages/koharu/components/preferences/PreferenceFields.tsx` passes `min={0} max={1}` to an `<input type="number">`. Those attributes are constraint-validation hints; the browser does not clamp typed text, and the raw `Number(value)` is written straight to the stored configuration.

2. **The stored value is fatal.** `detection::Processor::new` (`crates/koharu-pipeline/src/stages/detection.rs`) uses `ensure!` to reject a threshold outside `0.0..=1.0`. That `Err` propagates out of `initialize`, which is `.expect()`ed in `crates/koharu-app/src/app.rs:135`, so a bad preference becomes a hard startup failure rather than a recoverable one.

The same path applies to `bubble_threshold` and `panel_threshold`, and to any non-finite value.

### Expected behavior

- An out-of-range stored threshold should fall back to the model's recommended value with a warning, not prevent the application from starting. A preference should never be able to lock the user out of the UI that edits it.
- The preferences field should clamp to its declared bounds so the invalid value is never persisted.

### Environment

- Koharu 0.66.9
- Windows 11 Pro 26200

---

*AI disclosure, per [CONTRIBUTING.md](https://github.com/mayocream/koharu/blob/main/CONTRIBUTING.md): I hit this crash in normal use; the diagnosis above was written with Claude Code. The reproduction and the accompanying PR's tests were run locally against 0.66.9.*

robrich82:
Fixes #911.

A detection threshold stored outside `0.0..=1.0` currently makes Koharu unopenable: `detection::Processor::new` rejects it, the error propagates out of `initialize`, and `crates/koharu-app/src/app.rs:135` turns it into a fatal startup dialog. Recovering means hand-editing `~/.koharu/config.toml`.

### Changes

**`crates/koharu-pipeline/src/stages/detection.rs`** — an out-of-range or non-finite threshold is reset to `None` with a `tracing::warn!`, so `Model::load` uses `recommended_thresholds()` for that label. A stored threshold is a preference; it should not be able to lock the user out of the UI that edits it. This sanitizes in memory only and does not rewrite the configuration file, so the invalid value stays on disk (now harmless) until preferences are next saved.

With the only fallible path gone, `Processor::new` returns `Self` rather than `Result<Self>`, matching the sibling `ocr` and `translation` processors; `stages/mod.rs` drops the corresponding `?`.

**`packages/koharu/components/preferences/PreferenceFields.tsx`** — `NumberField` clamps to its declared `min`/`max` and rejects non-finite input before the value reaches the stored configuration. The HTML `min`/`max` attributes are constraint-validation hints and do not restrict typed text, which is how `15` reached the config file in the first place. An empty field still means `null`, i.e. the model default.

### Tests

- `out_of_range_thresholds_fall_back_to_the_model_defaults` covers `15.0`, `NaN`, and a valid `0.55` that must survive: `cargo test -p koharu-pipeline --lib out_of_range`.
- `clamps a threshold typed past its bounds before saving it` types `15` into Text threshold and asserts `text_threshold: 1` reaches `savePreferences`. Verified to fail without the `PreferenceFields.tsx` change.
- Full `tests/components/editor-components.test.tsx` (25 tests), `cargo fmt -p koharu-pipeline -- --check`, and `oxlint` all pass. `oxfmt --check` flags 60 files repo-wide, none of them touched here.

---

*AI disclosure, per CONTRIBUTING.md: I hit this crash in normal use; this fix was written with Claude Code. Every test listed above was run locally on Windows 11 against 0.66.9.*

### DIFF
--- a/crates/koharu-pipeline/src/stages/detection.rs
+++ b/crates/koharu-pipeline/src/stages/detection.rs
@@ -5,7 +5,7 @@ use std::{
     sync::{Arc, Mutex},
 };
 
-use anyhow::{Context as _, Result, anyhow, bail, ensure};
+use anyhow::{Context as _, Result, anyhow, bail};
 use async_trait::async_trait;
 use image::{DynamicImage, GrayImage, ImageFormat, Luma, RgbImage};
 use imageproc::{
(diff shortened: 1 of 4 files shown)


mayocream (approved):
LGTM

PR #862, opened 2026-07-12
feat(textbox): add rotation support for textbox rendering

zachelnet:
## Summary
This PR adds rotation support for textboxes in the rendering pipeline.

## What changed
- Added `rotation` handling in textbox data/model.
- Applied rotation transforms during textbox rendering.
- Adjusted positioning/alignment behavior for rotated textboxes.
- Preserved existing rendering behavior when rotation is not set.

## User-visible behavior
- Textboxes can now be rendered with rotation.
- Textboxes without a rotation value render the same as before (no behavior change).

## Validation
- Verified textboxes without rotation still render as before.
- Tested rotated textboxes across different angles.
- Checked positioning/alignment behavior for rotated textboxes in representative cases.

## Demo
https://github.com/user-attachments/assets/88837c2c-7fe6-46ba-b579-976275c21d1b

## AI Disclosure
AI-assisted (DeepSeek V4).  
The final patch was reviewed and validated by the contributor before submission.

PR #914, opened 2026-08-14
[Feature] : Toggle font color, weight and width between auto and manual

Hams-01:
Added options to be able to override auto for font weight, width and font color.
Global font weight and width can now be set in typesetting preferences.
Screenshot 2026-08-14 192528

### DIFF
--- a/crates/koharu-renderer/src/config.rs
+++ b/crates/koharu-renderer/src/config.rs
@@ -5,12 +5,18 @@ use specta::Type;
 #[serde(default)]
 pub struct TypesettingConfig {
     pub font_families: Vec<String>,
+    pub force_border_width: Option<f32>,
+    pub force_font_color: Option<String>,
+    pub force_font_weight: Option<u16>,
 }
 
 impl Default for TypesettingConfig {
     fn default() -> Self {
         Self {
             font_families: vec!["CCWildWords".to_owned(), "Adobe 黑体 Std".to_owned()],
+            force_border_width: None,
+            force_font_color: None,
+            force_font_weight: None,
         }
     }
 }
--- a/packages/koharu/components/preferences/TypesettingPreferences.tsx
+++ b/packages/koharu/components/preferences/TypesettingPreferences.tsx
@@ -5,7 +5,11 @@ import { useMemo } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { FontPicker } from '@/components/controls/FontPicker'
-import { PreferencePage } from '@/components/preferences/PreferenceFields'
+import {
+  PreferencePage,
+  PreferenceRow,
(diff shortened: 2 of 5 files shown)


crates/koharu-renderer/src/config.rs
```diff
@@ -5,12 +5,18 @@ use specta::Type;
 #[serde(default)]
 pub struct TypesettingConfig {
     pub font_families: Vec<String>,
+    pub force_border_width: Option<f32>,
+    pub force_font_color: Option<String>,
```
mayocream, line 9: Typography already uses `[u8; 4]` for color. Could we store this the same way (or as a bool, since the control is only Force Black)? Then the renderer doesn’t need a hex parser.

packages/koharu/lib/protocol.ts
```diff
@@ -497,6 +497,9 @@ export type TranslationConfig = {
 
 export type TypesettingConfig = {
 	font_families?: string[],
+  force_border_width?: number | null;
+  force_font_color?: string | null;
```
mayocream, line 501: This file no longer exists on main — typesetting types live in generated `packages/bridge/src/protocol.ts`. After rebase, please don’t recreate or hand-edit it.

packages/koharu/components/preferences/TypesettingPreferences.tsx
```diff
@@ -133,6 +137,121 @@ export function TypesettingPreferences({
           </p>
         )}
       </section>
+
+      {/* Grouped Overrides Section */}
+      <PreferenceSection
+        title="Global Overrides"
```
mayocream, line 143: Could these labels go through `t()` like the font-fallback section above? Hardcoded English will skip the other locales.

mayocream (changes requested):
Thanks for the first PR — this is a nice idea.

Render-time overrides on `TypesettingConfig`, with the inspector locking those fields, fits the existing typesetting prefs. No conflict with #876 (bulk font-size on nodes) or #862 (rotation).

1. **Rebase onto current main.** `packages/koharu/lib/protocol.ts` was moved to generated `packages/bridge/src/protocol.ts` (`// generated by koharu; do not edit`). Keep the Rust fields on `TypesettingConfig` and let specta regenerate the TS types — please don’t hand-edit protocol.

2. **Color shouldn’t be a hex string.** `Typography.color` is already `[u8; 4]`, and the UI here is only “Force Black”. Storing `Option<String>` plus a hex parser in the renderer is a second color format. Use RGBA like typography, or a bool if black is the only override.

3. **i18n.** The new prefs copy (and the inspector “overridden” tooltip) is hardcoded English. The rest of this page uses `t('settings.typesetting.*')` — same for these strings, including locale files.

Issue #936, opened 2026-08-15
Brush cursor freezes when pressing alt


FawazTakahji:
On Windows 11 at least when the brush cursor is on screen and alt is pressed or held, the brush cursor will freeze and the system cursor will show up.
Switching to another page, pressing alt, or switching to another app window unfreezes it.

<img width="1920" height="1032" alt="Image" src="https://github.com/user-attachments/assets/21b866e4-d703-432e-bd69-da5f705a7a5a" />

PR #937, opened 2026-08-16
fix: brush freeze bug, Change brush size with alt + scroll wheel

FawazTakahji:
This pr does two things, fix the bug in #936 and allow the user to change brush size by holding alt and scrolling.
koharu_xhaC1G658a


mayocream (approved):
LGTM

Issue #447, opened 2026-04-14
Broken text borders on punctuation marks


MrZimuk:
<img width="133" height="213" alt="Image" src="https://github.com/user-attachments/assets/9c0a62c7-0ed2-4e2c-a300-e72185b131b4" />

<img width="114" height="50" alt="Image" src="https://github.com/user-attachments/assets/817c9604-7640-4f4f-a758-f738aead83cc" />

Hecsall:
Since this is probably a related or the same issue, I'm adding this here. (í'm using Koharu v. 0.70.6)

When working with large images (for example, large vertical manhwa pages such as 1800 × 16000 px), text borders don't seem to scale correctly.

As you can see in my screenshots, at 158% zoom the borders appear huge, but when zooming in further, at 1600% zoom, they are actually the correct size. It seems that the borders don't use the same scaling ratio as the actual text.

<img width="560" height="700" alt="Image" src="https://github.com/user-attachments/assets/f0d45933-e195-4e51-8167-4d6968409794" />

<img width="1027" height="786" alt="Image" src="https://github.com/user-attachments/assets/1cb64119-e8da-42d4-81af-3aa453667474" />

Hecsall:
@mayocream I can confirm the "stroke scaling" issue is now fixed in the latest version 0.75.16.

For the "dots circle" issue, it is _mostly_ fixed, but I noticed some little holes that appear when you set a big stroke relative to the text size.
I've investigated this with AI and it appears to be a hard limitation from the `kurbo` package used for this.

When increasing the stroke of letters or symbols that have an inner path, for example letters/symbols with holes inside (B, O, R, P, D, # etc...) if the stroke is wider than half of the actual "hole" it starts to wrap on itself. You can see this in the below screenshot (taken from Koharu 0.75.16), notice the B from BORDER 10, and the # on the same text block.

This is not something that can be easily solved with the current approach, but I think its a smaller issue compared to before, since it appears only on visibly big strokes.

I'll leave in this collapsed block what the AI told me, if it helps.

<details>

<summary>AI summary of this issue</summary>

**The actual cause**

[...]

PR #951, opened 2026-08-18
fix(editor): replace text centered stroke with outward path

Hecsall:
> Note: This fix was developed with the assistance of AI tools.

This PR aims to fix issue #447

I'm no Rust developer, so I would suggest taking a closer look at this PR changes. It works, passes tests, and generates correct outputs, but I lack full understanding of the whole project and all its parts.

> The following is an extract of the issue/fix summary from the AI tool I used.

The text stroke previously used a centered stroke (`Stroke::new(width*2.0)`, drawn under a plain fill): a shape that grows both inward and outward from the glyph outline by half the width each way. 
When the border width exceeds a small feature's own radius (a dot, or a thin letter stroke at large border widths), the inward side of that stroke wraps past the shape's center and leaves a genuine transparent hole — producing the detached "ring" you can see in the below screenshots

The fix replaces that centered stroke with **outward-only** path dilation, using Vello's existing `FontEmbolden`/`expand_path` mechanism (previously only used for synthetic bold):

- The border pass is now a normal `Fill` in the border color, with `embolden` set to the border width — this only pushes the outline _outward_, so it can never create an inward-facing hole, regardless of how large the border width gets relative to the glyph's curvature.
- Kept the earlier zoom-scaling fix (disable hinting for this pass) so the dilation amount still scales correctly with canvas zoom.
- Used `Join::Round` for the dilation joins to avoid sharp miter spikes at glyph corners with large border widths.
- Removed the now-fully-unused `PreparedGlyphStyle` enum/`Stroke` variant from `koharu-rasterizer`'s wire format (per the repo's no-backward-compatibility rule) and bumped `PREPARED_FRAME_MANIFEST_VERSION` since the format shape changed.

### Issue example:

**Before the fix**

dots-pre

**After the fix**

koharu_18-08-2026_23-03-24

### DIFF
--- a/crates/koharu-rasterizer/src/frame.rs
+++ b/crates/koharu-rasterizer/src/frame.rs
@@ -4,15 +4,15 @@ use std::{collections::HashMap, sync::Arc};
 
 use vello::{
     FontEmbolden, Glyph, Scene,
-    kurbo::{Affine, BezPath, Diagonal2, Join, Rect, Stroke},
+    kurbo::{Affine, BezPath, Diagonal2, Join, Rect},
     peniko::{Blob, Color, Fill, FontData, Mix},
 };
 
 use crate::{
     Bounds, CompositionCommand, Error, FillRule, LayerId, LayerKind, PathElement, PreparedContent,
-    PreparedElementFrame, PreparedFrame, PreparedFrameBundle, PreparedFrameManifest,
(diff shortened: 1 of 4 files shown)


mayocream (approved):
Looks correct! Thank you!

PR #959, opened 2026-08-20
feat: drag-and-drop reordering of layers

### LINKED ISSUE #601: Drag to reorder text boxes
It would be useful to be able to change the number order of the text boxes, ideally by click-and-dragging the entries in the bottom right or otherwise perhaps by clicking some up and down or plus and minus buttons. The number order determines the order in which the language models gets the lines, so if the order is wrong it can affect the quality of the translation even with a strong model, especially in cases where a long line is spread over multiple boxes. Besides that, being able to change the order would also make proofreading and reading from the app a bit more convenient in those cases.

liksunrice:
## Summary:
This PR adds drag-and-drop reordering for layers inside `LayerInspector` to allows users to freely reorder layers quicker than the current method of clicking the arrow buttons when hovering over the layer in the selection.

## Changes:
- Added various states and refs in `LayerInspector` to track hover position and scroll boundaries
- `ScrollArea` Listener for auto-scroll logic when attempting to drag a layer to a location outside of view

### DIFF
--- a/packages/koharu/components/editor/Inspector.tsx
+++ b/packages/koharu/components/editor/Inspector.tsx
@@ -906,7 +906,6 @@ function LayerRow({
       <div
         data-selected={selected}
         data-expanded={expanded}
-        className='min-w-0 overflow-hidden rounded-lg transition-colors duration-150 data-[selected=true]:bg-accent motion-reduce:transition-none'
         className={`min-w-0 overflow-hidden rounded-lg transition-colors duration-150 data-[selected=true]:bg-accent motion-reduce:transition-none ${
           dragOverId === layer.id && dropPos
             ? dropPos === 'before'
--- a/packages/koharu/components/editor/Inspector.tsx
+++ b/packages/koharu/components/editor/Inspector.tsx
@@ -887,6 +887,7 @@ function LayerRow({
           e.dataTransfer.setData('text/plain', layer.id)
         }
         onDragStart(layer.id)
+      }}
       onDragEnd={(e) => {
         onDragEnd(e)
       }}
--- a/packages/koharu/components/editor/Inspector.tsx
+++ b/packages/koharu/components/editor/Inspector.tsx
@@ -475,6 +475,43 @@ function normalizeFontName(value: string): string {
   return value.normalize('NFKC').trim().replace(/\s+/g, ' ').toLowerCase()
 }
 
+function isDescendant(layers: Layer[], parentId: EntityId, childId: EntityId): boolean {
+  let current = layers.find((l) => l.id === childId)
+  while (current && current.parent) {
+    if (current.parent === parentId) return true
+    const parent = current.parent
+    current = layers.find((l) => l.id === parent)
+  }
+  return false
+}
+
+function isValidDrop(
+  layers: Layer[],
+  draggedId: EntityId,
+  targetId: EntityId,
+  pos: 'before' | 'after' | 'inside',
+  pageId: EntityId,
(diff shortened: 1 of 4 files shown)


mayocream (commented):
This seem overcomplicated to me, we should use external library to replace the hand-written drag-and-drop sorting: https://dndkit.com/concepts/sortable/

packages/koharu/components/editor/Inspector.tsx
```diff
@@ -73,6 +73,19 @@ import {
...
+  useDroppable,
+  DragEndEvent,
+  DragMoveEvent,
+  DragStartEvent,
+  PointerSensor,
+  useSensor,
+  useSensors,
+  DragOverlay,
+  pointerWithin,
+} from '@dnd-kit/core'
```
mayocream: Let's use `@dnd-kit/react` and use https://dndkit.com/react/hooks/use-sortable/

packages/koharu/components/editor/Inspector.tsx
```diff
@@ -500,21 +554,177 @@ function LayersInspector() {
...
+      setDragOverId(null)
+      setDropPos(null)
+      return
+    }
+
+    setDragOverId((prev) => (prev === overId ? prev : overId))
+    setDropPos((prev) => (prev === pos ? prev : pos))
+  }
+
+  const handleDragEnd = (event: DragEndEvent) => {
```
mayocream, line 650: This will still call `handleDrop` on a canceled drag (`event.canceled`, e.g. Escape), because `dragOverId`/`dropPos` stay set from the last hover.

Please gate on `!event.canceled` and use `event.operation.target` / `event.operation.position` here instead of React state. dnd-kit already gives you the pointer position, so the window `pointermove` listener can go.

packages/koharu/components/editor/Inspector.tsx
```diff
@@ -639,26 +887,66 @@ function LayerRow({
...
+  isDragOver: boolean
+  activeDropPos: 'before' | 'after' | 'inside' | null
+  isOverlay?: boolean
 }) {
   const { t } = useTranslation()
   const name = localizedLayerName(layer, index, t)
   const detail = localizedLayerKind(layer, t)
   const Icon = layerIcon(layer)
+
+  const { ref: sortableRef, isDragging: sortableIsDragging } = useSortable({
```
mayocream, line 899: This `ref` wraps the expanded `LayerEditor` as well, so dragging in a textarea or the opacity slider starts a layer drag. `useSortable` has `handle` / `handleRef` — please attach that to the row header only.

Same for the overlay: `expanded={expandedLayer === draggedId}` will clone the editor into the preview. Force the overlay to the header.

mayocream (changes requested):
Thanks for switching this to `@dnd-kit/react` and `useSortable`.

Two real issues before this can land:

1. `onDragEnd` still commits from React `dragOverId` / `dropPos` and does not check `event.canceled`. Escape, or releasing after the last valid hover, can still call `moveLayer`. Please commit only when the drag is not canceled, and read the target/position from the event (`event.operation.target`, `event.operation.position`). The window `pointermove` listener is not needed.

2. `useSortable` is on the whole row, so the expanded `LayerEditor` (source/translation fields, opacity slider) is part of the drag source — and the overlay can render that expanded editor too. Please attach the handle to the row header only, and keep the overlay as just the header.

The `moveLayer` parent/index mapping and the text-group drop rules look consistent with the existing inspector.

PR #977, opened 2026-08-22
Feat: Add keyboard navigation

### LINKED ISSUE #895: Use arrow keys to move through pages/menus
use arrow keys to advance through pages or text boxes when in appropriate side bars

nekojin116:
-Adds keyboard navigation to switch pages with left and right arrow
-Adds keyboard navigation for fonts list with instant font application so user preview the font 

https://github.com/user-attachments/assets/58d13e60-6e85-4182-b662-6ab83ea2f6c0

### DIFF
--- a/packages/koharu/components/controls/FontPicker.tsx
+++ b/packages/koharu/components/controls/FontPicker.tsx
@@ -3,7 +3,16 @@
 import { observeElementRect, useVirtualizer } from '@tanstack/react-virtual'
 import type { TFunction } from 'i18next'
 import { ChevronDown, ListFilter, Search, X } from 'lucide-react'
-import { useEffect, useMemo, useRef, useState } from 'react'
+import {
+  useEffect,
+  useImperativeHandle,
+  useMemo,
+  useRef,
+  useState,
+  forwardRef,
+  type KeyboardEvent,
+  type RefObject,
+} from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { useFontPreview } from '@/lib/queries'
@@ -45,6 +54,10 @@ type Filters = {
   useCase: string
 }
 
+interface FontListHandle {
+  focusRelative: (direction: -1 | 1) => void
+}
+
 const emptyFilters: Filters = { source: '', script: '', category: '', useCase: '' }
 
 export function FontPicker({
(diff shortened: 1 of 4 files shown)


packages/koharu/components/editor/CanvasWorkspace.tsx
```diff
@@ -124,6 +137,33 @@ export function CanvasWorkspace() {
       ? page?.layers.find((layer) => layer.id === selected[0] && layer.type === 'raster')
       : undefined
 
+  const moveToPage = useCallback(
+    (nextPage: PageSummary) => {
```
mayocream, line 141: This is a second page-switch path next to `PageRail.select`, but it never updates `selectedPages` / `selectedLayers`, and it doesn't write the prepared page into `projectKey`/`pageKey` when the canvas already swapped. Those are the bits that keep the rail, overlays, and layer shortcuts in sync — please follow that same sequence.

packages/koharu/components/editor/CanvasWorkspace.tsx
```diff
@@ -356,6 +396,19 @@ export function CanvasWorkspace() {
         requestCanvasFit()
         return
       }
+      if (event.key === 'ArrowLeft' || event.key === 'ArrowRight') {
```
mayocream, line 399: This listener only skips inputs, so Left/Right still page-switch while a slider (or other arrow-key control) is focused. Please ignore the event when it's already been handled, or when focus is on a control.

packages/koharu/components/controls/FontPicker.tsx
```diff
@@ -109,6 +124,7 @@ export function FontPicker({
     [orderedFamilies, value],
   )
   const activeFilterCount = Object.values(filters).filter(Boolean).length
+  const handleFocus = (family: string) => onChange(family)
```
mayocream, line 127: Callers treat `onChange` as a commit — inspector `setTypography`, settings font stack. Firing it from `onFocus` writes every font you arrow onto, and closing without selecting still leaves the last one. Keep commit on select.

mayocream (changes requested):
Thanks for this — keyboard page switching and font-list arrows are welcome.

A few real issues before this can land:

Page arrows go through a new `moveToPage` that doesn't follow `PageRail.select`. After a switch the rail still shows the old page, layer selection stays on the previous page, and query data lags the canvas until IPC returns. Please use the same selection + optimistic cache updates that path already does, rather than a second, slimmer switch.

Left/Right also fire for focused sliders and other non-input controls (zoom, opacity, open menus). The existing `editable()` check doesn't cover those, so the shortcut hijacks them. Skip when the event was already handled.

Font list `onFocus` calls `onChange`, which the inspector and typesetting settings treat as a commit (`setTypography` / writing the font stack). Arrowing then persists every row you pass, including the add-family picker. Preview is fine; commit should stay on select.

Happy to re-review after that.

PR #988, opened 2026-08-26
feat: add Hayai OCR model

NopeNopeGuy:
## Problem
Since existing OCR models are too slow or don't support the majority languages, add [Hayai OCR v2](https://huggingface.co/JustANormalTinkerer/hayai-ocr-v2) (~150M params, fast JA/ZH/KO/EN crop OCR) as an OCR stage option.

## Solution
 - New crates/koharu-ml/src/hayai_ocr/ port pinned to `hayai-ocr-v2@4a4ce47`: SigLIP2 NaFlex encoder (host patchification, per-image interpolated position embeddings) + GQA decoder with block-causal prefill, 2D mRoPE over patches / 1D over text, KV-cache greedy decoding matching upstream generate.
 - Registered as "hayai-ocr" across pipeline config, OCR stage, run CLI, preferences UI, README, and regenerated protocol.ts.
Intentional divergences (noted in code as well): greedy only, non-antialiased bilinear resize, native-GQA SDPA.

## Checks

- `cargo test -p koharu-ml --lib hayai` — processor unit tests pass; resize budget expectations verified against a direct replication of upstream's binary search (including the 4×4 → 256×256 upscale case).
- Structured output comparison against the authoritative implementation (pinned upstream Python, greedy num_beams=1, identical inputs):
   * benches/fixtures/ocr/title.png → both produce 対策委員会です! 
   * upstream README example 01.png (550×132) → both produce 知らない世界で見つけたイメージを
- `cargo test -p koharu-pipeline --lib` — 53 passed.
- `cargo clippy -p koharu-ml` clean for the new code; `git diff --check` clean.
- CPU device (LibTorch 2.12.1); no timing baseline recorded since no performance change is claimed relative to existing models.
- Used the APP to verify the changes work as well on test images.

## AI Disclosure

AI was used to generate the majority of this PR. However, all tests were run by a human, and the pipeline has been verified in app by a human as well.


crates/koharu-ml/src/hayai_ocr/model.rs
```diff
@@ -0,0 +1,807 @@
...
+            ],
+            1,
+        );
+
+        let m_vision = hidden.size()[1] - 1;
+        let total_length = m_vision + 1;
+        // Block-causal prefill mask: bidirectional across visual tokens, causal
+        // onto the text token (upstream `generate_block_causal_mask`).
+        let text_future =
+            Tensor::full([total_length, 1], f64::NEG_INFINITY, (kind, device)).triu(1);
```
mayocream: [P1] Build the blocked BOS column explicitly. `text_future` has shape `(total_length, 1)`, so `.triu(1)` zeros every entry: there is no element above the first superdiagonal in a one-column matrix. The resulting prefill mask is therefore entirely zero, letting every visual query attend to BOS. Upstream sets `mask[:m_vision, m_vision:] = False`, so this changes the visual states from layer 1 onward and can change OCR output. Please make the first `m_vision` entries of this column `-inf` and only the final BOS row zero (or derive the full upstream mask).

crates/koharu-ml/src/hayai_ocr/model.rs
```diff
@@ -0,0 +1,807 @@
...
+        }
+        let mut tokens = vec![first_token];
+
+        for step in 1..MAX_NEW_TOKENS {
+            let current_input = *tokens.last().expect("a token was generated");
+            seen_ids.push(current_input);
+            let embedding = self
+                .decoder
+                .token_embeddings
+                .forward(&Tensor::from_slice(&[current_input]).view([1, 1]));
```
mayocream: [P1] Move generated token IDs to the model device before the embedding lookup. `Tensor::from_slice` creates this tensor on CPU, while `token_embeddings.ws` is on CUDA, ROCm, or MPS when acceleration is selected. The BOS path above does the required `.to_device(pixel_values.device())`, but every cached decoding step omits it, so accelerator inference fails as soon as a second token is needed. Please construct this ID tensor on `pixel_values.device()` (and use that same device for any future index tensors).

mayocream (changes requested):
Two generation-path blockers need correction before merge: the prefill mask currently does not enforce block-causal attention, and cached decoding creates token IDs on CPU for accelerator-resident embeddings. I left both findings inline on the current head.

mayocream (approved):
Addressed both generation-path findings in 95aa2951: the prefill mask now explicitly blocks visual queries from the BOS column, and cached decoding moves token IDs to the model device. Focused Hayai tests pass.

PR #989, opened 2026-08-26
ColorWell: Track history of last 8 selected colors

ZenoArrows:
Allows you to easily pick between colors used previously

color-history


mayocream (dismissed):
Looks great, thank you!

The history store sits cleanly next to ColorWell, the canvas eyedropper only records when it doesn't already go through `set()`, and #861 is a different picker so no overlap.

PR #999, opened 2026-08-28
fix(ui): select CJK glyph variants from the interface language

StellarDragon:
## Summary

The desktop UI previously loaded separate Noto Sans JP, SC, and TC web fonts and selected regional Han glyphs by changing the font-family order. This update instead self-hosts the full [Noto Sans CJK variable OpenType collection](https://github.com/googlefonts/noto-cjk/raw/main/Sans/Variable/OTC/NotoSansCJK-VF.otf.ttc) through `next/font/local` and selects the appropriate OpenType glyph variants from the interface language.

### Example (Lang: Simplified Chinese)

image

## Changes

- Replace the three `next/font/google` imports with one local Noto Sans CJK variable font covering weights 100–900.
- Apply the local family to the desktop chrome through one CSS variable.
- Use [`font-variant-east-asian`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-variant-east-asian) for the language-specific variants:
  - `ja-JP`: `jis04`
  - `zh-CN`: `simplified`
  - `zh-TW`: `traditional`
- Preserve Korean localization through the font collection's `locl` forms selected by `lang`; CSS has no Korean-specific East Asian variant keyword.

The upstream file is an OTF-flavored TrueType Collection. It is stored with an `.otf` suffix because `next/font/local` accepts OpenType sources but rejects the `.ttc` filename extension; the bundled bytes are unchanged from the linked source.

## User-visible behavior

- Japanese, Simplified Chinese, Traditional Chinese, and Korean interface copy render from the same complete Noto Sans CJK collection.
- Han glyph forms follow the current `html lang` value set during language changes.
- Non-CJK interface languages keep the existing Japanese-default CJK forms.
- Canvas typesetting is unchanged; this only affects the desktop chrome.

## Verification

- `bun run --filter @koharu/app build`
- Confirmed the optimized export emits one local 32.7 MB OpenType asset with the 100–900 weight range and the three language selectors.
- Confirmed headless Chrome loads the collection and renders Japanese, Simplified Chinese, Traditional Chinese, and Korean language-tagged text.


mayocream (changes requested):
Let's use https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/font-variant-east-asian instead.

mayocream (approved):

PR #1007, opened 2026-08-29
feat(translation): add prompt controls and output validators

spgsroot:
## Why

Batch manga translation needs stronger guarantees than a target-language selector alone provides:

- The source language was optional in translation requests, so it could be serialized as `null` and leave both the prompt and provider APIs to infer it. An explicit source language makes translation direction deterministic and gives every backend the same language context.
- Additional instructions can refine Koharu's built-in prompt, but they cannot support workflows that need a completely different system contract. A system-prompt override enables those workflows without weakening the default prompt.
- LLMs occasionally leak untranslated source-script text even when instructed not to. This is especially costly in batch jobs because a single missed segment may survive until final review. Deterministic output validation catches these failures immediately and gives model-backed providers one bounded chance to regenerate the complete JSON response with precise corrective feedback.

## What changed

- Make `source_language` part of translation configuration and every `TranslationRequest`, and send it through prompts, DeepL, and Google Cloud Translation.
- Add a configurable system-prompt override while retaining separate additional instructions.
- Add reusable regex validation rules for translated `text` fields.
- Compile and validate rules when preferences or pipeline processors are created.
- Retry model-backed translations once with a full-response regeneration instruction; return an explicit validation error when the retry still fails or the provider cannot consume corrective prompts.
- Add UI controls in Translation settings and the runtime output picker.
- Add presets for Latin letters, Japanese kana, Chinese Han characters, and Korean characters, plus custom rules.
- Regenerate the TypeScript protocol and localize the new controls.

## Configuration change

`translation.source_language`, `translation.system_prompt`, and `translation.validators` are now part of the pipeline configuration. Existing local configurations need an explicit source language and validator list, for example:

```toml
[pipeline.translation]
source_language = "ja-JP"
target_language = "ru-RU"
validators = []
```

## Verification

- `cargo test -p koharu-translator --lib` — 54 passed
- `cargo test -p koharu-pipeline config::tests` — 8 passed
- `cargo check -p koharu-app`
- `bun run test tests/components/editor-components.test.tsx` — 37 passed
- `bun x tsc --noEmit`
- `bun run lint`
- `bun run build` — release desktop build completed
- Desktop CEF smoke test verified source/target selectors, system prompt, all four presets, and creation of the `[A-Za-z]` rule


mayocream (dismissed):
Thanks for this — it fits the current translator, pipeline, and UI patterns well.

Source language, the system-prompt override, and the regex validators all sit on the existing `TranslationConfig` / `TranslationRequest` path instead of a parallel one. Compiling rules on save/processor setup and giving model-backed providers a single bounded retry (hard fail for DeepL/Google/Caiyun) is the right shape.

Existing `config.toml` files should keep loading through koharu-config's default merge, so users should not need to hand-edit `source_language` / `validators` after upgrading.

Looks good to me.

PR #1024, opened 2026-08-31
feat(rendering): Added global overrides for font color, font weight and font border width

Hams-01:
Summary
Font Border width, color and weight can now be toggled between auto and a custom value.

Issue
Found border width to be unpredictable on small, unconventional text like sfx. Black text with a thin white margin is acceptable on most backgrounds.

Screenshot 2026-08-31 174549


mayocream (changes requested):
We should fix the auto color detection & border detection instead of set a global overrides.

PR #1041, opened 2026-09-05
feat(activity): name the page a pipeline run is working on

### LINKED ISSUE #1018: name the page a pipeline run is working on
## Problem

The pipeline already reports which page each stage is running against, and the job carries it, but the activity panel shows only the stage and the model. During a multi-page run there is no way to tell which page is being worked on; the only sense of position is the percentage.

## Proposal

Show the page label beside the model on the activity row's detail line.

```text
Ocr
Page 1 · manga-ocr                     40%
```

instead of `manga-ocr` alone.

Labels come from the pages query the rail has already loaded, so this resolves from cache rather than issuing a request, and the query stays disabled for jobs that have no page.

- No backend or protocol change: the page is already on the job.
- No additional request for the label.
- Export jobs are unaffected; they carry no page and keep their existing counter.
- Degrades gracefully: if the page is not in cache, or the stage reports no model, whichever half exists is shown without a stray separator.

## Product scope questions

Per the contributing guide I am opening this before sending a PR.

1. Is the page label wanted in the activity row at all, or do you prefer it to stay stage and model only?
2. Beside the model on one line as proposed, or on its own line?
3. The row is narrow and truncates. Should a long page label be elided in favour of the model, or the other way round?

lightglitch:
## Summary

During a multi-page run the activity panel showed only the stage and the model, so the only sense of position was the percentage. The pipeline already reports which page each stage is running against and the job carries it, so the label was available and simply not shown.

The detail line now reads `Page 1 · manga-ocr` instead of `manga-ocr`.

Labels come from the pages query the rail has already loaded, so this resolves from cache rather than issuing a request, and the query stays disabled for a job with no page. Joining with `filter(Boolean)` means an uncached label or a stage with no model shows whichever half exists, without a stray separator. No backend or protocol change.

## Related issue

Closes #1018, opened first per the contributing guide.

That issue asked three scope questions, which this answers one way — the label beside the model on the existing detail line, sharing its truncation. Happy to move it to its own line or change what gets elided if you'd prefer.

One correction to the issue text: its bullet about export jobs keeping their counter does not apply here. That came from a fork where export runs as a job; upstream has no export jobs in the job system, so there is nothing to leave unaffected.

## Verification

`bun run --filter @koharu/app test` (88 passed), `lint`, and `tsc --noEmit` are clean.

The existing activity-panel test gains an assertion that the row names the page rather than the model alone.

Not verified: the truncation behaviour of a long page label in a narrow panel was reasoned about, not measured against real labels.

## AI assistance

AI assisted in implementing this change and writing the description. Reviewed and tested by a human before submission.

### DIFF
--- a/packages/koharu/components/editor/ActivityCenter.tsx
+++ b/packages/koharu/components/editor/ActivityCenter.tsx
@@ -4,6 +4,7 @@ import { CircleAlert, Download, Square, X } from 'lucide-react'
 import { useTranslation } from 'react-i18next'
 
 import { call } from '@/lib/backend'
+import { usePages } from '@/lib/queries'
 import { useKoharuStore } from '@/lib/store'
 import { commands, type Download as DownloadState, type Job } from '@koharu/bridge/protocol'
 import { Button } from '@koharu/ui/components/button'
@@ -76,6 +77,16 @@ function DownloadGroup({ downloads }: { downloads: DownloadState[] }) {
 function JobItem({ job }: { job: Job }) {
   const { t } = useTranslation()
   const dismiss = useKoharuStore((state) => state.dismissJob)
+  // The pipeline reports which page it is on; the rail has already loaded the
+  // labels, so this resolves from cache rather than fetching.
+  const pages = usePages(job.page !== null).data
+  // Numbered by position in the rail, so the row says where in the run this
+  // is as well as which file: a scan filename alone rarely tells you.
+  const pageIndex = job.page ? (pages?.findIndex((page) => page.id === job.page) ?? -1) : -1
+  const page = pageIndex >= 0 ? pages?.[pageIndex] : undefined
+  const pageLabel = page
+    ? t('activity.pageLabel', { number: pageIndex + 1, label: page.label })
+    : undefined
   if (job.state === 'failed') {
     return (
       <Failure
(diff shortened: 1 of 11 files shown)


packages/koharu/components/editor/ActivityCenter.tsx
```diff
@@ -95,7 +100,9 @@ function JobItem({ job }: { job: Job }) {
               ? t(`phase.${job.stage}`, { defaultValue: job.stage })
               : t('activity.processing')}
           </span>
-          <p className='mt-0.5 truncate text-[10px] text-muted-foreground'>{job.model}</p>
+          <p className='mt-0.5 truncate text-[10px] text-muted-foreground'>
+            {[pageLabel, job.model].filter(Boolean).join(' · ')}
```
mayocream, line 104: Can you show an example screenshot where a page label is very long?

lightglitch: humm, I will test it.

lightglitch: Updated the PR with a multi-line solution.
Screenshot 2026-09-11 233055
Screenshot 2026-09-11 233045

mayocream: Hmmm, the actual page number being truncated, what do you think if only show the page number (index+1)? or only show last part of page label?

lightglitch: We can put the index at the beginning of the label like:

Page [index]: label 

and truncate the label like it is now.

mayocream (approved):
Thank you for the contribution!

PR #1044, opened 2026-09-05
fix(app): disable devtools to prevent F12 crash (#1042)

### LINKED ISSUE #1042: Pressing F12 (DevTools/console) crashes the app
### Description
Opening the developer console with **F12** crashes the app.

Reported by `@mig15_1947` in Mayo Lab Discord (#feedback):
https://discord.com/channels/1287227752470286367/1287227754105798663/1545808697954017360

### Steps to reproduce
1. Launch Koharu
2. Press `F12` (open console / DevTools)
3. App crashes

### Expected
DevTools/console opens without crashing (or F12 is ignored if DevTools are not meant to be user-facing).

### Actual
App crashes when F12 is pressed.

### Additional context
- Report language: English
- Discord message: "pressing F12 (opening console) will crash the app"
- OS / Koharu version / logs not provided yet — please attach logs from the Koharu log folder if you can reproduce:
  - Windows: `%LOCALAPPDATA%\Koharu\logs`
  - macOS: `~/Library/Logs/Koharu`
  - Linux: `~/.local/share/Koharu/logs` (or `$XDG_DATA_HOME/Koharu/logs`)

visheshgubrani:
## Summary

Pressing F12 (or Ctrl+Shift+I, Cmd+Opt+I) opened the internal CEF DevTools
window, and closing that window crashed Koharu. The failure is inside
 `tauri-runtime-cef`, the DevTools browser reuses the main webview's client, so
its lifecycle handler reports the main webview as closed when DevTools closes,
and the runtime then tears down the only window and the app exits.

Koharu does not expose DevTools to end users, so this change disables devtools
at the webview level and through the cargo feature:

- Set `devtools: false` on the main window in `tauri.conf.json` and
  `tauri.macos.conf.json`. The macOS config re declares the window list, so it
  needs the flag too. With the attribute disabled, tauri-runtime-cef installs
  its pre key guard, which swallows F12, Ctrl+Shift+I and Cmd+Opt+I before
  Chromium handles 
I didn't tease. I didn't start slow. them.
- Removed the `devtools` feature from the tauri dependency in `Cargo.toml`, so
  release builds cannot turn internal devtools back on through the feature
  flag. Nothing in the repository uses the devtools API surface.

Debug builds keep the CEF remote debugging endpoint at
`http://127.0.0.1:4000`, which is a command line switch and does not depend on
the devtools feature, so inspection with a browser or chrome-devtools-mcp is
unaffected.

The underlying close misrouting bug still exists in `tauri-runtime-cef`, disabling devtools removes the entry point that
triggers it.

## Related issue

Fixes #1042

## Verification

- `cargo check -p koharu -p koharu-app` passes.
- Linux (debug build, `bun dev`):
  - F12 and Ctrl+Shift+I no longer open DevTools and no longer crash the app.
  - Opening and closing DevTools is no longer possible from the window, the
    app stays running and closes normally through its own close path.
  - `curl http://127.0.0.1:4000/json/version` responds and the DevTools
    frontend is reachable through the remote debugging endpoint in a browser.
- Not verified: Windows and macOS behavior (no local environment) and a
  release profile build. The guard logic in `tauri-runtime-cef` covers all three
  platforms, but platform confirmation is still pending.

## AI assistance

AI assistance was used to trace the crash path in the pinned
`tauri-runtime-cef`revision and to review the fix plan

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,7 +128,6 @@ tauri = { git = "https://github.com/tauri-apps/tauri", rev = "6f34826405b5908cd5
     "macos-private-api",
     "specta",
     "x11",
-    "devtools",
 ] }
 tauri-build = { git = "https://github.com/tauri-apps/tauri", rev = "6f34826405b5908cd5c6fea7f38e33badb69ad54" }
 tauri-plugin-log = { git = "https://github.com/tauri-apps/plugins-workspace", rev = "1423992771e0b57582ca3b06a6adc46ec26aa784" }


crates/koharu/tauri.conf.json
```diff
@@ -24,7 +24,8 @@
         "title": "Koharu",
         "decorations": false,
         "transparent": false,
-        "visible": false
+        "visible": false,
+        "devtools": false
```
mayocream: This change is unnecessary, let's avoid changing it.

crates/koharu/tauri.macos.conf.json
```diff
@@ -17,7 +17,8 @@
         "trafficLightPosition": { "x": 16, "y": 23 },
         "hiddenTitle": true,
         "transparent": false,
-        "visible": false
+        "visible": false,
+        "devtools": false
```
mayocream: This change is unnecessary, let's avoid changing it.

mayocream (approved):

PR #1087, opened 2026-09-11
feat(translator): add OrcaRouter provider with API-key and PKCE login

kuswardhanietidims-svg:
## What

Adds OrcaRouter as a first-class model provider, plus a one-click connect flow so users can either paste an existing API key or authorize with their own OrcaRouter account.

- Provider entry in `crates/koharu-translator/src/provider.rs` (the `define_providers!` registry) — `base_url = https://api.orcarouter.ai/v1`
- Request mapping and discovery in `crates/koharu-translator/src/remote/orcarouter/`
- API-key choice: provider id `orcarouter`, label `OrcaRouter - API`, stored in the project's existing OS credential service via `koharu_secrets` under the `orcarouter` key (the same slot `Settings → Providers` already writes with `CredentialInput`)
- Connect flow: OAuth 2.0 + PKCE, **Flow A (loopback redirect)**, chosen because Koharu is a desktop application whose own process can bind `127.0.0.1`, so the code returns automatically and the user clicks once

**Affiliation disclosure:** I'm an engineer on the OrcaRouter team. This change is made on behalf of, and in coordination with, OrcaRouter.

## About OrcaRouter

[OrcaRouter](https://www.orcarouter.ai) is an OpenAI-compatible AI gateway built for both models and agents, with adaptive routing, automatic failover, zero-markup inference, observability, guardrails, and agent-tool governance.

It also runs gateway-level, zero-trust security for AI agents on the same endpoint — screening every prompt/response and governing every tool call on a default-deny basis, with no application code changes.

Discord: discord.gg/YEubt8enRA · X: https://x.com/OrcaRouter

## How the credential works

The key belongs to the user, not to this project: it is billed to their OrcaRouter account, listed in their console, and revocable at any time. No client secret is involved — PKCE binds the auth code to this process, so an intercepted code cannot be redeemed by anyone else.

Both ways in run through one seam. `CredentialProvider` has two adapters — `ApiKeyAdapter` (reads the pasted key) and `AuthorizationAdapter` (runs the browser flow) — and both return the same `Credential`. The provider, the model catalog and every AI entry point consume that type and cannot tell which was used. Tests assert exactly that.

Key lifecycle points:

- A PKCE-issued key is **durable, not refreshable**. No refresh grant is invented and no proactive refresh is scheduled; the key is reused until OrcaRouter revokes it.
- A relay `401` marks the **exact credential generation** that made the rejected request `needsReauth`. A late failure from an older request cannot disable a newer credential, and the old secret is never deleted before a successful replacement.
- The granted `scope` is read back from the exchange response rather than assumed from the request.
- Two origins, never derived from one another: auth is `https://www.orcarouter.ai` (authorize `/auth`, exchange `/api/v1/auth/keys`), inference and discovery are `https://api.orcarouter.ai/v1`. `ORCA_BASE_URL` covers a shared self-hosted origin; `ORCA_AUTH_BASE_URL` / `ORCA_API_BASE_URL` override each independently and explicit settings win. Remote origins require HTTPS; loopback may use plain HTTP.

## AI input entry points covered

Koharu has one hosted-inference seam; every entry point that can reach a provider was traced and wired:

| Entry point | File | Status |
| --- | --- | --- |
| Translation pipeline (with optional page image) | `crates/koharu-translator/src/remote/mod.rs` | wired |
| Editor processing selector | `packages/koharu/components/editor/InferenceControl.tsx` | wired, capability-filtered |
| Settings → Translation model selector | `packages/koharu/components/preferences/TranslationPreferences.tsx` | wired, capability-filtered |
| Settings → Providers (credential entry) | `ProviderPreferences.tsx` + `OrcaRouterConnection.tsx` | both auth methods |

Not covered, deliberately:

- **Koharu Agent** (`crates/koharu-agent`) is the one other LLM caller, but it is hard-wired to the Codex backend (`chatgpt.com/backend-api/codex/responses`) with its own OAuth device flow and does not consume `ProvidersConfig` at all. Routing it through OrcaRouter would be a separate architectural change, not a provider registration.
- Koharu has no embedding, image-generation, video or rerank entry point. The capability layer still filters for them so a future caller gets a correct list.

## Model discovery and capability filtering

Options come from the live catalog at `GET {api_base}/v1/models`. The key stays in the Rust process; the UI only ever receives model metadata (`Model {provider, model, name, vision, reasoning}`), which is the boundary Koharu already uses for every provider.

Filtering, by entry point:

- **Text**: `?capability=chat`, keeping only `supported_endpoint_types` that include `openai`, `anthropic`, `gemini` or `openai-response`, and excluding `image-generation`, `openai-video` and `jina-rerank`.
- **Multimodal**: the chat set, narrowed to records whose `architecture.input_modalities` explicitly contains the modality the entry point uploads. With **Vision** on (Koharu's default) only models declaring `image` are offered; an undeclared modality fails closed, so a record with no `architecture` block never appears in the multimodal list.
- **Embedding** / **image generation** / **video** / **rerank**: matched strictly by their own endpoint type; no name-based guessing.

These options are bound to the real selector as a filtered prop, not merely guarded before sending. When Vision is toggled or the provider changes the list is recomputed, and a saved model that no longer qualifies is dropped rather than silently kept. Discovery is bounded (20s timeout, 8 MiB body, 2000 models, endpoint-type allowlist). On failure a small verified seed is served and labelled `degraded` — never a free-text field and never an unverified example list.

### Verified fallback seed

Every id below was confirmed present in the live catalog on **2026-09-11** (`GET https://api.orcarouter.ai/v1/models?capability=chat`, 161 models returned): `openai/gpt-5.5`, `anthropic/claude-opus-4.8`, `google/gemini-3.5-flash`, `deepseek/deepseek-v4-pro`, `orcarouter/auto`. Context length and input modalities are carried from the catalog response. **No reasoning-effort ladder is claimed**: `/v1/models` publishes no reasoning field and no `supported_parameters`, so the seed deliberately keeps `reasoning: false` rather than inventing `low`/`medium`/`high`/`xhigh` from model names. If OrcaRouter starts advertising reasoning support, that should be added from the response, not guessed here.

## Provider evidence

- Inference API (OpenAI-compatible, `/v1/chat/completions`) and model list: `https://api.orcarouter.ai/v1`; `GET /v1/models` requires the Bearer key and returns the models the workspace can actually call.
- Authorization endpoints and the loopback/out-of-band rules: `https://www.orcarouter.ai/auth` and `https://www.orcarouter.ai/api/v1/auth/keys`.
- Consent screen, revocation and account management: `https://www.orcarouter.ai/console/authorized-apps` revokes every key issued to this app in one click.
- Model namespace follows `vendor/model` and is preserved verbatim.
- Maintenance owner and verification date: OrcaRouter engineering team (I am an engineer on the team, see affiliation disclosure above); endpoints re-verified 2026-09-11 against the live service.

## UI

`Settings → Providers` shows the OrcaRouter row with both choices side by side: a masked `OrcaRouter - API` credential field with its clear control, and `Connect with OrcaRouter`, which starts the browser authorization and shows the consent URL for browsers that do not open automatically. Both are labelled distinctly wherever they appear. Labels resolve through all nine shipped locale catalogs (`en-US`, `zh-CN`, `zh-TW`, `ja-JP`, `ko-KR`, `ru-RU`, `es-ES`, `pt-BR`, `tr-TR`), and the locale key-parity test passes.

Screenshots below are real: headless Chromium driving the repository's own `ProviderPreferences` and `ModelPicker` components, whose model list is served by a local backend that holds the key. The browser never receives a credential.

1. `Settings → Providers` — both authentication entries visible, key control masked.
2. Text state — the live chat catalog (161 models), dropdown open.
3. With an image attachment / Vision on — the filtered list (122 models that declare image input).

## Testing

**UI suite** — `bun run --filter @koharu/app test`: **108 passed / 0 failed / 0 skipped** across 14 files. Baseline on unmodified `main` in this container was 89 passed across 12 files, so this adds 19 tests and regresses none.

New tests: `tests/lib/provider-capabilities.test.ts` (9) and `tests/components/orcarouter-connection.test.tsx` (10), covering: both auth choices registered and independently labelled; the API-key path never starts a PKCE login; clearing a stored key; the authorization running through the connect command; denial releasing the busy state; `pagehide` clearing busy/hint and allowing a second login without remounting; explicit cancel; a stale response from an invalidated attempt not surfacing; `needsReauth` reported without any refresh claim; and capability filtering — image attachment leaving only declared-vision chat models, the current text model being invalidated, an undeclared architecture failing closed, and other providers' behaviour unchanged.

**Rust** — `cargo fmt --check` clean on every touched file. The provider, PKCE and catalog modules were compiled and their tests executed for real against the repository's source (see *Verification environment* below): **34 passed / 0 failed**, covering RFC 7636 Appendix B challenge vector, fresh verifier/state per attempt, unpadded base64url, verifier and key absent from `Debug` output, authorize URL on the auth origin with S256 and no verifier in the URL, explicit-override precedence over the shared base, HTTPS enforcement with loopback exemption, the chat endpoint preserving the `/v1` segment, capability filters per endpoint type, multimodal fail-closed, degraded-seed metadata, 401 marking only the rejected generation with no refresh, reauthorization clearing `needsReauth`, and login-lock release after a denial.

Separately, an end-to-end harness exercised the real `pkce.rs` against a local fake authorization server: **8 passed / 0 failed** — loopback code delivery and a closeable response page, state-mismatch rejection, denial, timeout, `POST /api/v1/auth/keys` with the exact `{code, code_verifier, code_challenge_method: S256}` body, scope read back from the response, 400/403/429 classified terminally with no credential material in the error, and a transport failure not being mistaken for an issued key.

**Live checks with a real key**

- `GET https://api.orcarouter.ai/v1/models?capability=chat` via the new discovery code: **161 models**, not degraded.
- `GET ...?capability=chat` filtered to declared image input: **122 models**, every one declaring `image`.
- `POST https://api.orcarouter.ai/v1/chat/completions` with `orcarouter/auto` and a strict `json_schema` response format: HTTP 200, valid JSON returned — the request shape this provider sends is accepted.
- These live runs depend on `ORCAROUTER_API_KEY`; the API key is read from the environment, never printed, never committed and never passed to the browser.

**Lint / format / typecheck** — `@koharu/app lint`, `@koharu/bridge lint` and `tsc --noEmit` on the bridge all pass.

Known limitations, stated plainly:

- `cargo test --workspace` (the CI command) was **not** run. The touched crate depends on `koharu-ml` → `koharu-torch` / `koharu-llama` / `koharu-diffusion`, which need an external libtorch (CUDA) and libclang; that toolchain is not present here, and no amount of waiting makes it present. Instead the affected modules were compiled and their tests run against the repository's own source through a harness that supplies minimal stand-ins for the sibling crate surface (`koharu_secrets` in-memory store, a smaller `Model`/`Error`/`prompt`/`send_json` surface, and a `display_name` that matches the real one). This is why the harness exists, and it is the strongest verification available in this environment.
- `bun run check` (oxfmt) reports 61 files unformatted on `main` in this container, before my change; every file I touched is formatted and passes `oxfmt --check`.
- The browser authorization itself was driven against a local fake authorization server, not against the real consent screen: approving a real consent screen requires a human, and no worker may forge that step. The real endpoints, paths, parameter set and body are asserted from the protocol specification.

## Verification environment

`bun` and a Rust toolchain are not installed in the container this was prepared in; both were obtained to run the repository's own commands (`bun install --frozen-lockfile --ignore-scripts`, `bun run --filter @koharu/app test`, `cargo`). All commands above were executed; the only gate that could not be run is the full `cargo test --workspace`.

## Notes for the maintainer

- This touches credentials and where prompts are sent, so it will likely trigger the security-review gate. I have not applied any label to my own PR; a maintainer review at the exact head commit is what is needed.
- `packages/bridge/src/protocol.ts` is generated by specta from the Rust types. It is hand-updated here to match the new commands and config types, because regenerating requires building `koharu-app` (same libtorch limitation). The Rust definitions it mirrors are in `commands/preferences.rs`.


mayocream (changes requested):
> a one-click connect flow so users can either paste an existing API key or authorize with their own OrcaRouter account.

Please only add API key workflow, authorize workflow doesn't add any value.

PR #1088, opened 2026-09-11
chore(app): default the processing scope to the selection

### LINKED ISSUE #1086: Make selection the default scope for the pipeline
### Description

Currently the page scope is somewhat redundant: selecting a single page and running with selection scope is equivalent. Moreover, if you have a selection, you probably want to run with selection scope. I suggest making selection scope the default, or even removing page scope entirely,

### Additional context

_No response_

spgsroot:
## Summary

The runtime selector above the canvas now starts on the selected-pages scope instead of the current page. The page rail always auto-selects the active page, so the default run still processes exactly the current page, while an explicit page-rail selection is now honored by default instead of being silently ignored in favor of the active page.

Scope removal (the stronger option in the issue) is intentionally left out; this is the conservative change and keeps the current-page escape hatch.

Closes #1086

## Test plan

- Updated the store test mock and the runtime-selector tests to the new default (`/Scope Selection/`), and renamed the default-run test to `runs the selected pages by default`; the default run still sends `{ scope: 'pages', value: ['page'] }`. The processing guide notes the new default.
- `bun run --filter @koharu/app test` — 89/89 passed on latest main.
- `bun run --filter '@koharu/*' lint` — clean in all packages.

AI assistance: Claude Code implemented and verified this change; the human author reviewed the complete diff.

### DIFF
--- a/packages/docs/en/guides/processing.mdx
+++ b/packages/docs/en/guides/processing.mdx
@@ -4,7 +4,7 @@ description: 'Run detection, OCR, translation, and inpainting at the scope your 
 icon: 'play'
 ---
 
-The processing selector above the canvas controls the page scope and stages. Start with the smallest scope that covers the work. The selector starts on the selected-pages scope, which follows the page-rail selection.
+The processing selector above the canvas controls the page scope and stages. Start with the smallest scope that covers the work.
 
 ## Choose a scope
 
--- a/packages/docs/en/guides/processing.mdx
+++ b/packages/docs/en/guides/processing.mdx
@@ -4,7 +4,7 @@ description: 'Run detection, OCR, translation, and inpainting at the scope your 
 icon: 'play'
 ---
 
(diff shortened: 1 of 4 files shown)


packages/docs/en/guides/processing.mdx
```diff
@@ -4,7 +4,7 @@ description: 'Run detection, OCR, translation, and inpainting at the scope your
 icon: 'play'
 ---
 
-The processing selector above the canvas controls the page scope and stages. Start with the smallest scope that covers the work.
+The processing selector above the canvas controls the page scope and stages. Start with the smallest scope that covers the work. The selector starts on the selected-pages scope, which follows the page-rail selection.
```
mayocream: Let's do not change this line.

spgsroot: Done — reverted the docs line in 3ddf0878.

mayocream (approved):
LGTM

PR #1093, opened 2026-09-12
Fix long project names pushing the delete button off screen

FawazTakahji:
## Summary

Long project names pushed the delete button off screen. Fixed by keeping the name from expanding past its container.

before:
image

after:
image

the dots look like that because of the font (noto-sans-cjk), we can cutoff the text instead of you want:
image

## Test plan

I looked at it 👍


mayocream (approved):
Thank you for catching it!

PR #1095, opened 2026-09-12
feat: export a project as a CBZ archive

### LINKED ISSUE #1016: export a project as a CBZ archive
## Problem

Koharu imports CBZ but cannot export it. `docs/reference/formats-and-data.md` lists CBZ under page import only, so shipping a finished chapter means exporting loose PNGs and zipping them by hand.

There is prior art. #471 added CBZ and CB7 export in April. It was reviewed with "Please hold this; I will review it after redesigning the data model", then closed unmerged after it went stale. The redesign has since landed, so a fresh implementation against the current scene model is possible.

Export also has no progress reporting today: `export_pages` runs inside the command future, so the invoke promise stays pending for the whole run and nothing appears in the activity panel. That is tolerable for one page and much less so for a whole project.

## Proposal

Add CBZ alongside PNG and PSD, under a `File -> Export` submenu.

- Choosing CBZ asks which format the pages inside the archive use: PNG, JPEG or WebP. All three are already import formats, so an exported archive opens again as a project.
- Members follow the comic archive convention, `P001`, `P002`, matching #471. Numbering widens past three digits when a project needs it, so members still sort.
- Quality for the lossy formats lives in a new Export section in settings.
- Export registers a job and returns immediately, so the activity panel shows progress and can stop it.
- A stopped or failed export removes its partial archive rather than leaving an unreadable file.

## Constraints found while prototyping

These shaped the proposal and may be worth knowing before discussing scope.

- **JPEG cannot carry alpha.** Its encoder rejects RGBA input outright rather than dropping the channel, so pages have to be composed onto a matte first. White matches the PSD preview.
- **Members are best stored uncompressed.** Every supported page format is already compressed, so deflating again costs time for no meaningful size reduction.
- **AVIF is export-only in this build.** It encodes through `ravif`, but `image` reads AVIF only through its `avif-native` feature and the `dav1d` C library, so an AVIF archive would export and then fail to re-import. Left out for that reason.
- **CB7 has the same asymmetry.** Koharu cannot import `.cb7`, so unlike #471 it is not included here.

- An exported archive re-imports into Koharu with page order preserved.
- Members are named by position, independent of page labels, which are not ordered and usually already carry their own number.
- Loose PNG and PSD export behaviour is unchanged.
- Export progress is visible and stoppable.
- No new runtime dependency: `zip` is already a direct dependency of `koharu-app`.

## Product scope questions

A previous implementation was submitted as #1009 and closed because this product behavior had not been agreed with maintainers first. I am opening this issue before requesting review again.

1. Is CBZ export wanted, and is a `File -> Export` submenu the right home for it?
2. Should CBZ follow the page rail selection the way PNG and PSD do, or always cover the whole project? A single-page comic archive seems rarely useful, so the prototype always exported the project and said so in the menu label.
3. Is `P001` the naming you want, following #471, while loose files keep the existing `0001_page-label` convention?
4. Should CB7 be included for parity with #471? If so, should `.cb7` import come with it so the formats stay symmetric?
5. Should the page format be chosen per export, as proposed, or fixed in settings?
6. Is wiring export into the job system acceptable as part of this change, or would you prefer it as a separate one?

lightglitch:
## Summary

Koharu imports CBZ but could not write one, so shipping a finished chapter meant exporting loose PNGs and zipping them by hand.

CBZ joins PNG and PSD under a `File -> Export` submenu. Choosing it asks how the pages inside are encoded — PNG, JPEG or WebP, each already an import format, so an exported archive opens again as a project. Members are named by position, `P001` onward, widening past three digits when a project needs it; loose PNG and PSD keep their existing `0001_page-label` names. A CBZ always covers the whole project rather than the rail selection, since a single-page comic archive is rarely useful, and the menu label says so.

Export reports progress now, which it did not before: `export_pages` ran inside the command future, so the invoke promise stayed pending for the whole run and nothing appeared in the activity panel. It registers a job and returns, so the panel names the page it has reached and how many are written, the way a pipeline run does, and can stop it. A stopped or failed export removes its partial archive rather than leaving an unreadable file.

Jobs carry a `kind` so the panel can describe an export instead of naming a pipeline stage. Exports are tracked in their own map because they read a snapshot, so they neither block a pipeline run nor are blocked by one; `stop_job` resolves both.

Two encoding details are not obvious from the diff. JPEG cannot carry alpha and its encoder rejects RGBA outright rather than dropping the channel, so pages are composed onto white first, matching the PSD preview. Members are stored rather than deflated, because every supported page format is already compressed.

Quality lives with the encoding that uses it — PNG lossless, JPEG 90, WebP 85 — rather than in settings. Those sit above the usual photographic defaults because screentone and fine line work compress badly.

## Related issue

Closes #1016, where this was agreed before implementation, including the job-kind change.

Two scope questions from that issue are answered differently from the original proposal, both in the direction of less machinery:

- **Quality is fixed, not a setting.** No Export section, no config file, no clamping of hand-edited values.
- **AVIF and CB7 remain out.** Koharu can encode AVIF but cannot decode it without `dav1d`, and cannot import `.cb7` at all, so either would export and then fail to re-import.

## Verification

`cargo fmt --check`, `cargo clippy -p koharu-app --all-targets -- -D warnings`, `cargo test -p koharu-app --lib output::` (8 passed), `bun run --filter @koharu/app test` (96 passed), `tsc --noEmit`, and `bun run --filter @koharu/app lint` are all clean. `packages/bridge/src/protocol.ts` is regenerated, not hand-edited.

New tests cover each encoding producing its own container and decoding back, the JPEG matte, `P001` member naming and its widening, loose files keeping their label, an archive preserving member order with members stored intact, and the activity row naming an export and the page it has reached.

Not verified: the save dialog, the stop path removing a partial archive, and the JPEG matte against a real page were reasoned about and unit tested, but I could not exercise the built application end to end on this machine.

## AI assistance

AI assisted in implementing this change and drafting this description. Reviewed and tested by a human before submission.


crates/koharu-app/src/commands/processing.rs
```diff
@@ -61,7 +70,11 @@ pub enum JobState {
 
 #[derive(Default)]
 pub(crate) struct Processing {
+    /// Pipeline runs, which are mutually exclusive with each other.
     pub(crate) stops: Mutex<HashMap<JobId, StopToken>>,
+    /// Export runs. An export reads a snapshot, so it neither blocks a
+    /// pipeline run nor is blocked by one.
+    pub(crate) exports: Mutex<HashMap<JobId, StopToken>>,
```
mayocream: this duplicates with stop, please remove it

packages/bridge/src/protocol.ts
```diff
@@ -182,7 +182,7 @@ export type Error = string;
 
 export type Event = { type: "started"; run: RunId } | { type: "text_delta"; run: RunId; delta: string } | { type: "reasoning_delta"; run: RunId; delta: string } | { type: "tool_started"; run: RunId; call_id: string; name: string } | { type: "tool_finished"; run: RunId; call_id: string; name: string; changed: boolean; output: string } | { type: "completed"; run: RunId; message: string } | { type: "failed"; run: RunId; message: string } | { type: "cancelled"; run: RunId };
 
-export type ExportFormat = "png" | "psd";
+export type ExportFormat = "png" | "psd" | { cbz: ImageEncoding };
```
mayocream: let's only use png inside the cbz

packages/bridge/src/protocol.ts
```diff
@@ -268,6 +274,7 @@ export type InpaintingModel = { model: "lama" } | { model: "aot-inpainting" } |
 
 export type Job = {
 	id: JobId,
+	kind: JobKind,
```
mayocream: we don't need a separate job kind, this should be in export command

packages/koharu/components/app/ExportCbzDialog.tsx
```diff
@@ -0,0 +1,82 @@
...
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+} from '@koharu/ui/components/dialog'
+import { Label } from '@koharu/ui/components/label'
+import { RadioGroup, RadioGroupItem } from '@koharu/ui/components/radio-group'
+
+/// Every option is also an import format, so an exported archive opens
+/// again as a project.
+const encodings: readonly ImageEncoding[] = ['png', 'jpeg', 'webp']
```
mayocream: we don't need user's action

packages/koharu/components/editor/ActivityCenter.tsx
```diff
@@ -90,7 +91,9 @@ function JobItem({ job }: { job: Job }) {
   if (job.state === 'failed') {
     return (
       <Failure
-        message={job.error || t('activity.processingFailed')}
+        message={
+          job.error || t(exporting ? 'activity.exportFailed' : 'activity.processingFailed')
```
mayocream: exporting should not handled separately

mayocream (changes requested):

crates/koharu-app/src/commands/processing.rs
lightglitch: Removed. The export job goes into `stops`, and `stop_job` already finds it there, so `processing.rs` is now unchanged.

One consequence worth naming: `process` refuses while `stops` is non-empty, so an export and a pipeline run are now mutually exclusive.

packages/bridge/src/protocol.ts
lightglitch: Done. Pages inside the archive are PNG, and `ImageEncoding` is gone along with JPEG, WebP, their quality values and the white matte JPEG needed.

packages/bridge/src/protocol.ts
lightglitch: Removed. An export is a plain job now; the export command owns it.

packages/koharu/components/app/ExportCbzDialog.tsx
lightglitch: Deleted. The menu item starts the export directly.

packages/koharu/components/editor/ActivityCenter.tsx
lightglitch: Reverted; the file is identical to main again.

The row therefore reads "Processing" during an export, since it has no stage to name. Happy to change that if you would rather it said otherwise.

mayocream (approved):

PR #1096, opened 2026-09-13
feat(navigator): say how many pages a selection covers

### LINKED ISSUE #1017: act on a page selection from the rail, and say what it covers
## Problem

Selecting several pages in the page rail gives no running total. The selection is visible only as highlighted rows, so on a long rail there is no way to tell how many pages are selected without counting them.

The same information is missing where it matters most. `Export PNG` and `Export PSD` act on the selection, falling back to the active page, but the menu entries do not say how many pages they will write. A selection made in the rail becomes invisible at the moment of exporting, which is exactly when its size matters.

Acting on that selection is also a round trip. Processing or deleting pages means leaving the rail for a menu, even though the pages are already under the pointer and already selected.

## Proposal

Make the rail say what is selected, and let it act on the selection.

**The selection is stated where it is made**

- The header shows `N selected` above the filter once more than one page is selected.
- The `PNG` and `PSD` export entries show the same count.

**The rail can act on pages**

- A process button in the header runs the same command as `Process -> Selected Pages` when more than one page is selected, and `Process -> Project` otherwise.
- A delete button beside it deletes the selection, or every page when there is no multi-selection. It asks first, naming the count, since it is bulk and destructive and sits next to a button that starts a pipeline run.
- Each page row reveals a process button on hover that runs only that page.

Scope belongs to the header and a row means the page it sits on. Both header buttons follow the selection the same way, so the two read alike, and their tooltips and accessible names say which of the two they will do before anything happens.

A single selection is left unannotated: it is the ordinary case, and it behaves the same as acting on the active page, so a `1 page` badge would only add noise.

- Counts are derived from the value the commands actually receive, so what a control says cannot disagree with what it does.
- The rail indicator is announced to assistive technology when it changes.
- The process buttons are unavailable while a run is in flight, since the backend refuses a second one.
- Deleting asks for confirmation and names what it will remove. It stays undoable, as page deletion already is.
- Nothing changes when zero or one page is selected, beyond the buttons meaning the whole project.
- No backend or protocol change: every command already exists.

## Product scope questions

A previous implementation of the count alone was submitted as #1010 and closed because this product behavior had not been agreed with maintainers first. I am opening this issue before requesting review again, and have widened it because the count and the buttons are the same question: what the rail knows about a selection and what it lets you do with it.

1. Is a selection count wanted in the rail header, or would you rather it live in the status bar?
2. Should the export entries carry the count, or is the rail indicator enough on its own?
3. Are process and delete buttons wanted in the rail at all, or should those stay in the menus?
4. Is following the selection the right behaviour for the header buttons, or should they always mean the whole project?
5. Should the header hold both buttons, or is delete too destructive to sit one icon away from a pipeline run?
6. Is "more than one" the right threshold throughout?

Screenshot in a comment below.

lightglitch:
## Summary

Selecting several pages in the rail gave no running total. The selection showed only as highlighted rows, so on a long rail there was no way to tell how many pages it held without counting them.

The same number was missing where it matters most. `Export PNG` and `Export PSD` act on the selection, falling back to the active page, but said nothing about how many pages they would write, so a selection made in the rail became invisible at the moment of exporting.

The rail header now states the count above the filter.

- The rail states it in a `role="status"` region, so it reaches assistive technology when it changes rather than only being visible.
- One selected page stays unannotated. It is the ordinary case and behaves exactly like acting on the active page, so a `1 page` badge would only add noise.

Frontend only: one component, nine locale files, one test file. No backend, no protocol change, no new commands.

## Related issue

Closes #1017.

That issue proposes the count **and** process/delete buttons in the rail. This is the count alone, which is the half agreed in the issue thread — the buttons are deliberately not here, so the design question about action buttons living in the rail stays open and can be answered separately.

## Verification

`bun run --filter @koharu/app test` (97 passed), `tsc --noEmit`, and `bun run --filter @koharu/app lint` are clean.

One test added: the rail announcing `3 selected` through a status region while staying silent at one selection.

Not verified: this was unit tested rather than exercised in the built application, so the visual placement of the count in a narrow rail has not been seen on screen.

## AI assistance

AI assisted in implementing this change and drafting this description. Reviewed and tested by a human before submission.

### DIFF
--- a/packages/koharu/components/editor/PageRail.tsx
+++ b/packages/koharu/components/editor/PageRail.tsx
@@ -238,6 +238,17 @@ export function PageRail() {
               {pages.length}
             </span>
           </div>
+          {/* One selected page behaves exactly like acting on the active
+              page, so saying so would be noise. */}
+          {selected.length > 1 && (
+            <span
+              role='status'
+              aria-live='polite'
+              className='ml-auto text-[9px] text-muted-foreground tabular-nums'
+            >
+              {t('navigator.selected', { count: selected.length })}
+            </span>
+          )}
         </header>
 
         {importing && (
--- a/packages/koharu/public/locales/en-US/translation.json
+++ b/packages/koharu/public/locales/en-US/translation.json
@@ -298,7 +298,8 @@
     "pages": "Pages",
     "rename": "Rename",
     "renameDescription": "Choose a new name for this page.",
-    "renameTitle": "Rename page"
+    "renameTitle": "Rename page",
+    "selected": "{{count}} selected"
   },
   "outputPicker": {
     "instructions": "Translation instructions",
--- a/packages/koharu/public/locales/es-ES/translation.json
+++ b/packages/koharu/public/locales/es-ES/translation.json
@@ -298,7 +298,8 @@
     "pages": "Páginas",
     "rename": "Cambiar nombre",
     "renameDescription": "Elige un nombre nuevo para esta página.",
-    "renameTitle": "Cambiar nombre de página"
+    "renameTitle": "Cambiar nombre de página",
+    "selected": "{{count}} seleccionadas"
   },
   "outputPicker": {
     "instructions": "Instrucciones de traducción",
--- a/packages/koharu/public/locales/ja-JP/translation.json
+++ b/packages/koharu/public/locales/ja-JP/translation.json
@@ -298,7 +298,8 @@
     "pages": "ページ",
     "rename": "名前を変更",
     "renameDescription": "このページの新しい名前を入力します。",
-    "renameTitle": "ページ名を変更"
+    "renameTitle": "ページ名を変更",
+    "selected": "{{count}}件選択中"
   },
   "outputPicker": {
     "instructions": "翻訳指示",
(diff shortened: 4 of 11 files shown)


packages/koharu/components/app/TitleBar.tsx
```diff
@@ -57,6 +57,10 @@ export function TitleBar() {
 
   const closeProject = () => void call(commands.closeProject).catch(() => undefined)
 
+  // Exactly what the export commands below receive, so the count in the menu
+  // cannot disagree with what is written.
+  const exportCount = exportSelection(selectedPages, page?.id).length
```
mayocream: export counter is duplicated with selected counter, let's remove it

mayocream (changes requested):

packages/koharu/components/app/TitleBar.tsx
lightglitch: But it's a different information, unless you want to join both counts?
image

mayocream: "5 selected" already means these will be processed, let's do not duplicate the status, we make the UX straightforward so we don't need redundant info.

lightglitch: Removed. The rail count stands on its own now, and `TitleBar.tsx` is unchanged from main.

You are right that it was the same fact twice.

mayocream (approved):

